PR #21, opened 2021-08-26
Simplified regex strings

andreidraga:
The ending square brackets shouldn't be escaped because they have a special meaning only if they match some starting square brackets (but this is not the case here). So, they are normal chars.


camilbancioiu:
Sorry, the code in `erdpy/ide` is deprecated and should not be modified.

PR #29, opened 2021-09-02
Extra config args

claudiu725:
- Use local erdpy.json from the current directory, if available
- When running erdpy from the command line, the config file is checked for any default arguments
  - Replaced usages of `sys.argv` with an extra argument `args` so that the default arguments can be added from the config file first
  - Changed argument order for `is_arg_present` for consistency - `args` first
  - Note: arguments provided in the command line can override those in the config files
- Changes regarding exception handling:
  - Exceptions no longer print the entire stack trace
  - Wrapped the `FileNotFound` exception when reading files, which simplifies the error message

### DIFF
--- a/erdpy/config.py
+++ b/erdpy/config.py
@@ -1,3 +1,5 @@
+from itertools import chain
+from logging import Logger
 import os.path
 from typing import Any, Dict
 
@@ -175,3 +177,27 @@ def write_file(data: Dict[str, Any]):
         utils.write_json_file(LOCAL_CONFIG_PATH, data)
     else:
         utils.write_json_file(CONFIG_PATH, data)
+
+
+def add_config_args(argv):
+    if len(argv) < 2:
+        return argv
+
+    func, subcommand, *_ = argv
+    config = read_file()
+    if func not in config:
+        return argv
+
+    extra_func = config[func]
+    if subcommand not in extra_func:
+        return argv
+
+    extra_args = []
+    for key, value in extra_func[subcommand].items():
+        extra_args.append(f'--{key}')
+        if value is True:
+            continue
+        extra_args.append(str(value))
+
+    print(f"Added extra args from erdpy.json: {argv + extra_args}")
+    return argv + extra_args
--- a/erdpy/config.py
+++ b/erdpy/config.py
@@ -1,7 +1,7 @@
 from itertools import chain
 from logging import Logger
 import os.path
-from typing import Any, Dict
+from typing import Any, Dict, List
 
 from erdpy import errors, utils
 
@@ -197,7 +197,11 @@ def add_config_args(argv):
         extra_args.append(f'--{key}')
         if value is True:
             continue
-        extra_args.append(str(value))
+        if isinstance(value, List):
+            for item in value:
+                extra_args.append(str(item))
+        else:
+            extra_args.append(str(value))
 
     # the verbose flag is an exception since it has to go before the command and subcommand
     # eg. erdpy --verbose contract deploy
--- a/erdpy/config.py
+++ b/erdpy/config.py
@@ -180,20 +180,20 @@ def write_file(data: Dict[str, Any]):
 
 
 def add_config_args(argv):
-    if len(argv) < 2:
+    try:
+        func, subcommand, *_ = argv
+    except ValueError:
         return argv
 
-    func, subcommand, *_ = argv
     config = read_file()
-    if func not in config:
-        return argv
 
-    extra_func = config[func]
-    if subcommand not in extra_func:
+    try:
+        config_args = config[func][subcommand]
+    except KeyError:
         return argv
 
     extra_args = []
-    for key, value in extra_func[subcommand].items():
+    for key, value in config_args.items():
         extra_args.append(f'--{key}')
         if value is True:
             continue
--- a/erdpy/config.py
+++ b/erdpy/config.py
@@ -4,6 +4,7 @@ from typing import Any, Dict
 from erdpy import errors, utils
 
 ROOT_FOLDER_NAME = "elrondsdk"
+LOCAL_CONFIG_PATH = os.path.join(os.getcwd(), "erdpy.json")
 CONFIG_PATH = os.path.expanduser("~/elrondsdk/erdpy.json")
 
 DEFAULT_GAS_PRICE = 1000000000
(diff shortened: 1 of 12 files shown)


catalinnnn (approved):

erdpy/config.py
```diff
@@ -162,10 +165,56 @@ def get_defaults() -> Dict[str, Any]:
...
+        return utils.read_json_file(CONFIG_PATH)
+    return dict()
 
 
 def write_file(data: Dict[str, Any]):
-    utils.write_json_file(CONFIG_PATH, data)
+    if os.path.isfile(LOCAL_CONFIG_PATH):
+        utils.write_json_file(LOCAL_CONFIG_PATH, data)
+    else:
+        utils.write_json_file(CONFIG_PATH, data)
```
camilbancioiu, line 179: Consider creating a new function `resolve_config_path()` which returns either `LOCAL_CONFIG_PATH` or `GLOBAL_CONFIG_PATH`.

erdpy/config.py
```diff
@@ -1,9 +1,12 @@
...
+from logging import Logger
 import os.path
-from typing import Any, Dict
+from typing import Any, Dict, List
 
 from erdpy import errors, utils
 
 ROOT_FOLDER_NAME = "elrondsdk"
+LOCAL_CONFIG_PATH = os.path.join(os.getcwd(), "erdpy.json")
 CONFIG_PATH = os.path.expanduser("~/elrondsdk/erdpy.json")
```
camilbancioiu, line 10: We should rename `CONFIG_PATH` to `GLOBAL_CONFIG_PATH` now that we have `LOCAL_CONFIG_PATH`.

erdpy/config.py
```diff
@@ -162,10 +165,56 @@ def get_defaults() -> Dict[str, Any]:
...
+        utils.write_json_file(CONFIG_PATH, data)
+
+
+def add_config_args(argv):
+    try:
+        func, subcommand, *_ = argv
+    except ValueError:
+        return argv
+
+    config = read_file()
```
camilbancioiu, line 188: Consider using `resolve_config_path()` and passing it as argument directly to `utils.read_json_file()` here, obviating the need for the function `read_file()` above.

erdpy/config.py
```diff
@@ -162,10 +165,56 @@ def get_defaults() -> Dict[str, Any]:
...
-    utils.write_json_file(CONFIG_PATH, data)
+    if os.path.isfile(LOCAL_CONFIG_PATH):
+        utils.write_json_file(LOCAL_CONFIG_PATH, data)
+    else:
+        utils.write_json_file(CONFIG_PATH, data)
+
+
+def add_config_args(argv):
+    try:
+        func, subcommand, *_ = argv
```
camilbancioiu, line 184: Consider renaming `func` to `command`, for consistency with `subcommand`.

erdpy/config.py
```diff
@@ -162,10 +165,56 @@ def get_defaults() -> Dict[str, Any]:
...
+        if key_arg in argv:
+            continue
+        extra_args.append(key_arg)
+        if value is True:
+            continue
+        if isinstance(value, List):
+            for item in value:
+                extra_args.append(str(item))
+        else:
+            extra_args.append(str(value))
```
camilbancioiu, line 208: A separate function called `determine_extra_args()` (or another name) could contain this code.

erdpy/config.py
```diff
@@ -162,10 +165,56 @@ def get_defaults() -> Dict[str, Any]:
...
+        config_args = config[func][subcommand]
+    except KeyError:
+        return argv
+
+    extra_args = []
+    for key, value in config_args.items():
+        key_arg = f'--{key}'
+        # arguments from the command line override the config
+        if key_arg in argv:
+            continue
```
camilbancioiu, line 200: What happens when the user specifies `--somekey` in the local config file, but then writes `--somekey=somevalue` on the command line? Will `argv` contain the entire `--somekey=somevalue` or will it split at the equal sign `=`?

If the elements of `argv` aren't already split at `=`, then the `if` will not reach `continue` and both `--somekey=somevalue` and `--somekey` will enter `final_args`.

erdpy/config.py
```diff
@@ -162,10 +165,56 @@ def get_defaults() -> Dict[str, Any]:
...
+
+    # the verbose flag is an exception since it has to go before the command and subcommand
+    # eg. erdpy --verbose contract deploy
+    verbose_flag = '--verbose'
+    pre_args = []
+    if verbose_flag in extra_args:
+        extra_args.remove(verbose_flag)
+        pre_args = [verbose_flag]
+
+    final_args = pre_args + argv + extra_args
```
camilbancioiu, line 218: Would it be possible to operate directly on the `args` object instead? It might be easier to simply inject the local config values into the `args` object instead of modifying `argv` before parsing.

Or better yet, if possible, `parser.set_defaults(...)` might read the values from the local config file?

camilbancioiu (changes requested):
Refactoring only (and some suggestions).

erdpy/config.py
claudiu725: Done.

erdpy/config.py
claudiu725: Done.

erdpy/config.py
claudiu725: This `read_file()` function is used in multiple places and it's also used outside of the `utils` module. I'm keeping this as-is.

erdpy/config.py
claudiu725: Done.

erdpy/config.py
claudiu725: Done. Created a `determine_final_args` function.

erdpy/config.py
claudiu725: Good catch!
I added handling for this.

erdpy/config.py
claudiu725: I tried using `set_defaults`, but this fails with something along the lines:
`erdpy contract COMMAND [-h] ... deploy: error: the following arguments are required: --pem, --keyfile, --ledger, --nonce, --gas-limit`
At the moment the creation of the parser is in a circular dependency with the arguments that are given from the command line. This is because we want some arguments to be required, but we may provide one of several (eg. we need a wallet which we can specify either with `--pem` or `--keyfile` and at least one of them is required). Because of this circular dependency the `set_defaults` doesn't work, as it adds the default arguments, but requires a parser to have been built in the first place.

This could (probably?) be rewritten by using [`ArgumentParse.add_mutually_exclusive_group`](https://docs.python.org/3/library/argparse.html#mutual-exclusion) which should eliminate the circular dependency.
Here, my first thought was to add the extra args handling but keep the rest of the code unchanged, in order to not break anything. The refactoring should be done in another PR probably.

camilbancioiu (approved):

catalinnnn (approved):

Issue #31, opened 2021-09-07
Json serialization bug [counter tutorial]


romaingauthier:
Querying the counter smart contract as described in the tutorial ([here](https://docs.elrond.com/developers/tutorials/counter/#interact-with-the-deployed-contract)):

```
erdpy contract query $CONTRACT_ADDRESS --function="get" --proxy="https://testnet-gateway.elrond.com"
```

yields 

```
[
    Traceback (most recent call last):
  File "/<HOME>/anaconda3/envs/elrond/bin/erdpy", line 8, in <module>
    sys.exit(main())
  File "/<HOME>/anaconda3/envs/elrond/lib/python3.8/site-packages/erdpy/cli.py", line 32, in main
    _do_main()
  File "/<HOME>/anaconda3/envs/elrond/lib/python3.8/site-packages/erdpy/cli.py", line 56, in _do_main
    args.func(args)
  File "/<HOME>/anaconda3/envs/elrond/lib/python3.8/site-packages/erdpy/cli_contracts.py", line 307, in query
    utils.dump_out_json(result)
  File "/<HOME>/anaconda3/envs/elrond/lib/python3.8/site-packages/erdpy/utils.py", line 139, in dump_out_json
    json.dump(data, outfile, indent=4, cls=ObjectEncoder)
  File "/<HOME>/anaconda3/envs/elrond/lib/python3.8/json/__init__.py", line 179, in dump
    for chunk in iterable:
  File "/<HOME>/anaconda3/envs/elrond/lib/python3.8/json/encoder.py", line 429, in _iterencode
    yield from _iterencode_list(o, _current_indent_level)
  File "/<HOME>/anaconda3/envs/elrond/lib/python3.8/json/encoder.py", line 325, in _iterencode_list
    yield from chunks
  File "/<HOME>/anaconda3/envs/elrond/lib/python3.8/json/encoder.py", line 438, in _iterencode
    o = _default(o)
  File "/<HOME>/anaconda3/envs/elrond/lib/python3.8/site-packages/erdpy/utils.py", line 34, in default
    return json.JSONEncoder.default(self, obj)
  File "/<HOME>/anaconda3/envs/elrond/lib/python3.8/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type QueryResult is not JSON serializable

MWFIAE:
I made a while pr 2 days ago for that :)
[...]

PR #33, opened 2021-09-08
Fix regression bug for contract query

MWFIAE:
Fixing the following error message that is thrown if a QueryResult object is getting serialized for display in the console.

File "/home/wagnerm/python3.9/lib/python3.8/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type QueryResult is not JSON serializable

It is caused because the following line only checks for inheritance of the Object class. 
https://github.com/ElrondNetwork/elrond-sdk-erdpy/blob/main/erdpy/utils.py#L32

Without this fix every call to `erdpy contract query ` fails.

### DIFF
--- a/erdpy/contracts.py
+++ b/erdpy/contracts.py
@@ -8,13 +8,14 @@ from erdpy import config, constants, errors
 from erdpy.accounts import Account, Address
 from erdpy.interfaces import IElrondProxy
 from erdpy.transactions import Transaction
+from erdpy.utils import Object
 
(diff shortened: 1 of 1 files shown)


MWFIAE:
Fixes #31

camilbancioiu (approved):
Excellent, thanks.

andreibancioiu (approved):

PR #42, opened 2021-09-20
Added ComputeRewardCheckpointEnableEpoch to config

MWFIAE:
Since the [enableEpochs.toml](https://github.com/ElrondNetwork/elrond-go/blob/master/cmd/node/config/enableEpochs.toml#L112) for the nodes was updated we need to update erdpy too.

Otherwise the user receives the following error message:
"ValueError: unrecognized configuration keys {'ComputeRewardCheckpointEnableEpoch'}"


MWFIAE:
Btw. I just used 5 because that's the value that was used in the node configuration, might be better to set it to 0 or 1, like most other values.

MWFIAE:
@camilbancioiu Would be great if you could take a look, more and more people are getting those errors :)

claudiu725 (approved):

camilbancioiu (approved):

PR #60, opened 2021-11-05
Update template creation logic to handle "meta"

MWFIAE:
Updated the template creation logic to also handle the new "meta" directory, that replaced the "abi" directory


MWFIAE:
Closed because it's already fixed by #59

PR #76, opened 2021-12-13
Refactor ledger signing + better exceptions handling

bogdan-rosianu:
- refactor signing (now the Account signs the transaction, instead of getting the private key from an account and signing with it elsewhere)
- better handling of Ledger exceptions (when the device isn't unlocked, or on Elrond app)
- add ledger signing for contracts as well


erdpy/accounts.py
```diff
@@ -76,11 +82,55 @@ def get_secret_key(self) -> bytes:
...
+        secret_key = self.get_secret_key()
+        signing_key: Any = nacl.signing.SigningKey(secret_key)
+
+        data_json = transaction.serialize()
+        signed = signing_key.sign(data_json)
+        signature = signed.signature
+        signature_hex = signature.hex()
+        assert isinstance(signature_hex, str)
+
+        return signature_hex
```
camilbancioiu, line 95: Consider the following instead:
```suggestion
        data_json = transaction.serialize()
        signed = signing_key.sign(data_json)
        signature = signed.signature
        assert isinstance(signature, bytes)
        return signature.hex()
```

erdpy/interfaces.py
```diff
@@ -32,6 +26,14 @@ def to_dictionary_as_inner(self) -> Dict[str, Any]:
         return {}
 
 
+class IAccount:
+    def get_secret_key(self) -> bytes:
```
camilbancioiu, line 30: Is `get_secret_key()` still necessary?

bogdan-rosianu: removed

erdpy/accounts.py
bogdan-rosianu: done

erdpy/cli_contracts.py
```diff
@@ -235,7 +250,10 @@ def _prepare_contract(args: Any) -> SmartContract:
 
 
 def _prepare_sender(args: Any) -> Account:
```
AdoAdoAdo, line 252: shouldn't the function return Union[Account, LedgerAccount] instead of Account ?

erdpy/transactions.py
```diff
@@ -54,16 +49,16 @@ def receiver_username_decoded(self) -> str:
...
-        bytes = self.__dict__.get(field, None).encode("utf-8")
-        encoded = base64.b64encode(bytes).decode()
+        field_bytes = self.__dict__.get(field, None).encode("utf-8")
+        encoded = base64.b64encode(field_bytes).decode()
         return encoded
 
     def _field_decoded(self, field: str) -> str:
         return base64.b64decode(self.__dict__.get(field, None)).decode()
 
     def sign(self, account: Account):
```
AdoAdoAdo, line 59: maybe use Union[Account, LedgerAccount] for the account?

AdoAdoAdo (changes requested):

erdpy/cli_contracts.py
bogdan-rosianu: removed the Union as now Account is the base class of the LedgerAccount

erdpy/transactions.py
bogdan-rosianu: removed the Union as now Account is the base class of the LedgerAccount

camilbancioiu (dismissed):

AdoAdoAdo (dismissed):

AdoAdoAdo (approved):

camilbancioiu (approved):

keyur555:
Hello, would it be possible to Release a new Version with this Changes? Currently, Ledger Deployments do not work because of the Error: CRITICAL:cli:No wallet provided.

Thank you! :)

bogdan-rosianu:
> Hello, would it be possible to Release a new Version with these Changes? Currently, Ledger Deployments do not work because of the Error: CRITICAL:cli:No wallet provided.
> 
> The Changes in here look like the fix this.
> 
> Thank you! :)

Hi @keyur555 ! A new release will follow soon

bogdan-rosianu:
@keyur555 erdpy v1.0.24 has been released

PR #90, opened 2022-01-25
Fix rate limit

claudiu725:
Github API is rate limited. This means that at some point the github actions CI builds raise exceptions, resulting in a failed build:
```
requests.exceptions.HTTPError: 403 Client Error: rate limit exceeded for url: https://api.github.com/repos/ElrondNetwork/elrond-wasm-rs/releases
```

Github encourages caching responses. The cached requests do not count towards the total rate limit.
https://docs.github.com/en/rest/overview/resources-in-the-rest-api#conditional-requests

I used the `requests-cache` python library, which does this properly - note the `cache_control=True` argument.
The `use_cache_dir=True` argument makes it so that the cache is saved in the user's cache folder `~/.cache/`. This way the requests are cached across multiple `erdpy contract new --template ...` calls. This is useful especially since the tests have 8 such calls in `test_cli_contracts.sh`.

### DIFF
--- a/.github/workflows/erdpy-up-macos.yml
+++ b/.github/workflows/erdpy-up-macos.yml
@@ -18,9 +18,10 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
+      - uses: rlespinasse/github-slug-action@v3.x
       - name: Install
         run: |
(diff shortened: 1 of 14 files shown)


andreibancioiu (approved):
:+1:

camilbancioiu (approved):

PR #93, opened 2022-02-01
Fix build (test) issues on Github Actions by setting erdpy's "github_api_token"

andreibancioiu:
New step in erdpy.yml: set `github_api_token`.

Using the token is safe because:
 - At the start of each workflow run, GitHub automatically creates a unique `GITHUB_TOKEN` secret.
 - The token's permissions are limited to the repository that contains your workflow.
 - The permissions are set to `restricted` at the repository level. Additionally, the workflow explicitly sets all possible (known as of February 2022) permissions to `none`. 

References:
 - https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret
 - https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

### DIFF
--- a/.github/workflows/erdpy.yml
+++ b/.github/workflows/erdpy.yml
@@ -32,8 +32,9 @@ jobs:
         pip3 install pytest
     - name: Set github_api_token
       run: |
-        mkdir ~/elrondsdk && touch ~/elrondsdk/erdpy.json
+        mkdir ~/elrondsdk
         export PYTHONPATH=.
+        python3 -m erdpy.cli config new test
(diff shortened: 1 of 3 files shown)


claudiu725 (approved):

.github/workflows/erdpy.yml
```diff
@@ -8,6 +8,25 @@ on:
     branches: [ main, development ]
   workflow_dispatch:
 
+# At the start of each workflow run, GitHub automatically creates a unique GITHUB_TOKEN secret (we will pass it below to erdpy config).
```
iulianpascalau, line 11: creates an* unique....

iulianpascalau (approved):
small typo

PR #102, opened 2022-02-04
Transaction simulation & cost simulation: fix & redesign

andreibancioiu:
- Include transaction cost simulations in _stdout / outfile_ when the `--simulate` flag is active.
- On local testnet, for proxy, use it's own `economics.toml`.
- Cleanup / remove deprecated `erdpy tx cost`.
- Refactor / redesign some parts of the Proxy-related components
- Refactor / redesign the CLI output (the stdout / outfile _dumps_) - tried to avoid breaking changes (with somehow partial success, see notes below).
- Optimize / refactor transaction awaiting (`--wait-result`).
- Improved some CLI tests.
- Added output examples in `--help`.
- Fix call to parse_validator_pem() - pass a Path object (partial, temporary fix for https://github.com/ElrondNetwork/elrond-sdk-erdpy/issues/75).

Breaking change on `erdpy tx get` (could not work around it in a _clean-enough_ way) - the fetched transaction is on the field `transactionOnNetwork` of the output; previously it was dumped as a single-level object.

In order to avoid breaking changes, the _stdout / outfile_ now contains both the new and the deprecated fields (duplication in the output).


erdpy/proxy/core.py
```diff
@@ -140,24 +143,19 @@ def get_hyperblock(self, key) -> Any:
...
-            if self.is_transaction_finalized(tx_hash):
-                return self.get_transaction(tx_hash=tx_hash, with_results=True)
-            time.sleep(1)
-        return dict()
-
-    def is_transaction_finalized(self, tx_hash):
-        last_nonce = self.get_last_block_nonce("metachain")
-        last_hyperblock = self.get_hyperblock(last_nonce)
-        finalized_transactions = last_hyperblock["transactions"]
+            time.sleep(AWAIT_TRANSACTION_PERIOD)
```
claudiu725, line 152: As it's written right now, this now waits for `100 * 5 = 500` seconds instead of 100, as the argument suggests.

claudiu725 (changes requested):

erdpy/proxy/core.py
andreibancioiu: :tada: as always, very very nice catch!

claudiu725 (approved):

camilbancioiu (approved):

PR #105, opened 2022-02-17
Extend argument input types

mateigraura:
Extend the functionality of cli:contract --argument field to support more types. Added the following options:
- bech32 address (`erd1qqq...`)
- boolean values
- ascii, such as token identifiers or names can be passed directly

Hex values are treated as before, so are numeric ones

### DIFF
--- a/erdpy/contracts.py
+++ b/erdpy/contracts.py
@@ -16,6 +16,7 @@ HEX_PREFIX = "0x"
 ERD_BECH32_PREFIX = "erd"
 FALSE_STR_LOWER = "false"
 TRUE_STR_LOWER = "true"
+STR_PREFIX = "str:"
 
 
 class QueryResult(Object):
@@ -31,7 +32,8 @@ class SmartContract:
         self.bytecode = bytecode
         self.metadata = metadata or CodeMetadata()
 
-    def deploy(self, owner: Account, arguments: List[Any], gas_price: int, gas_limit: int, value: int, chain: str, version: int) -> Transaction:
+    def deploy(self, owner: Account, arguments: List[Any], gas_price: int, gas_limit: int, value: int, chain: str,
+               version: int) -> Transaction:
         self.owner = owner
         self.compute_address()
 
@@ -73,7 +75,8 @@ class SmartContract:
         address = bytes([0] * 8) + bytes([5, 0]) + address[10:30] + owner_bytes[30:]
         self.address = Address(address)
 
-    def execute(self, caller: Account, function: str, arguments: List[str], gas_price: int, gas_limit: int, value: int, chain: str, version: int) -> Transaction:
+    def execute(self, caller: Account, function: str, arguments: List[str], gas_price: int, gas_limit: int, value: int,
+                chain: str, version: int) -> Transaction:
         self.caller = caller
 
         arguments = arguments or []
@@ -103,7 +106,8 @@ class SmartContract:
 
         return tx_data
 
-    def upgrade(self, owner: Account, arguments: List[Any], gas_price: int, gas_limit: int, value: int, chain: str, version: int) -> Transaction:
+    def upgrade(self, owner: Account, arguments: List[Any], gas_price: int, gas_limit: int, value: int, chain: str,
+                version: int) -> Transaction:
         self.owner = owner
 
         arguments = arguments or []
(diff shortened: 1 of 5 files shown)


erdpy/contracts.py
```diff
@@ -177,16 +180,33 @@ def _interpret_return_data(self, data: str) -> Any:
...
-    if as_string.startswith(HEX_PREFIX):
-        return _prepare_hexadecimal(as_string)
-
-    return _prepare_decimal(as_string)
+    as_str = str(argument)
+    as_hex = _to_hex(as_str)
+    return as_hex
+
+
+def _to_hex(arg: str):
```
andreibancioiu, line 188: Although this is indeed a big improvement and helps a lot for common scenarios, the assumptions that "true" / "false" are boolean values (and never plain strings), or "erd1..." is always an address (and never a string) could lead to some issues in the future.

Currently, erdpy doesn't handle contract ABIs and cannot do informed coding / decoding of arguments (as erdjs does).

However, until then, how do you feel about changing your implementation as follows?

```
prepare_argument(argument):
    if argument starts with "bool:", do top level encoding of a boolean (as you did in your implementation)
    else if argument starts with "address:", then do top level encoding of an address (as you did in your implementation)
    else if argument starts with "ascii:", then ...
    else if argument starts with "utf8:", then ...
    else if argument starts with "bytes:", then ... (assert hexadecimal input here)
    else if argument starts with "int:", then ...
    else if argument starts with "uint:", then ...
    (fallback for backwards compatibility) else if argument starts with HEX_PREFIX (as before, old implementation)
    (fallback for backwards compatibility) else if argument is numeric ...
```

Then, the user would be able to do as follows:

```
--arguments bool:true ascii:foobar bytes:ABBA uint:7 0x0BAD 42
```

Of course, we'll then deprecate the old (current) way in upcoming versions.

CC: @claudiu725 @dorin-iancu - how does this sound in your opinion (until we do proper ABI handling)?

References:
 - https://docs.elrond.com/developers/developer-reference/elrond-serialization-format/

claudiu725: I agree with adding prefixes - sometimes the arguments can be ambiguous, especially if we allow any kind of strings to be passed as arguments. However, if we require strings to have a prefix, then the other types can be implied safely.
The prefix for strings, which we already use in mandos scenarios is `str:`. With this, there's no ambiguity between `true` and `str:true`.
The arguments above would look like:
`--arguments true str:foobar 0xABBA 7 0x0BAD 42`

Note: Later on, in another PR, we could add support for concatenation and fixed-width and nested types, bringing this closer to the existing mandos scenario syntax (eg. `u8:42|nested:str:hello|biguint:42`)

mateigraura: I'm down to change the implementation @andreibancioiu. I thought the same as @claudiu725, adding the `str/ascii` prefix should be enough for this iteration.

andreibancioiu: :rocket: 

Extra reference (from @andrei-marinica), which might help as well (now or in future PRs):

https://docs.elrond.com/developers/mandos-reference/values-simple/

mateigraura: added the `str:` prefix.  saw that some tests failed. passing locally.

mateigraura: can be re-reviewed imo. mandos test failing in erdpy-build workflow. re tested locally and passes, please have a look and let me know if it's something i could do or it's workflow related.
local versions: `rustc:1.60.0-nightly`, `vm tools: v1.4.39`

claudiu725 (approved):

erdpy/contracts.py
claudiu725: The tests were failing because vm tools `v1.4.42` which was required by the templates' test was not released yet.
As it has been released today, the tests now pass.

It's ready for review.

erdpy/contracts.py
```diff
@@ -177,16 +181,33 @@ def _interpret_return_data(self, data: str) -> Any:
...
+    if arg.startswith(HEX_PREFIX):
+        return _prepare_hexadecimal(arg)
+
+    if arg.isnumeric():
+        return _prepare_decimal(arg)
+    elif arg.startswith(ERD_BECH32_PREFIX):
```
andreibancioiu, line 195: This will require `address:...` in the future, as well.

claudiu725: `address:` in the mandos format currently generates dummy addresses from simple strings. I would rather not mix them.
I think for now this PR is fine as it is.

mateigraura: cool, feel free to merge then :)

andreibancioiu: @mateigraura, not all commits are signed. Do you think you can redo the PR using signed commits?

If not possible, let us know - and we will temporarily deactivate the "all signed commits" constraint.

mateigraura: I'd rather not, would be best if this one gets merged so the discussion remains in the merged PRs history

iulianpascalau: Next PRs should contain only signed commits. Otherwise we won't merge them as it is a security risk.

PR #141, opened 2022-05-19
Update localnet

andreibancioiu:
- Update golang (used to compile the Node and the Proxy for localnet): **go1.17.6**
 - Do not automatically patch (edit) `enableEpochs.toml` anymore.
 - Allow one to overwrite Node configuration by placing files in folder called `localnet-node-config` (located whether in `~/elrondsdk` or **in the current directory**). Thus, for example, one can overwrite the default `enableEpochs.toml` with a customized one - though, this should not be necessary in **most cases**. **Note however that not all network features are active at epoch 0. For example, smart contract deployments become active at epoch 1.**
 - Change chain ID from `local-testnet` to `localnet`.
 - Allow one to specify the duration of an epoch, by adding an entry in `testnet.toml`, as follows:

```
[timing]
rounds_per_epoch = 30
```


claudiu725:
Tested locally - it works with:
```bash
erdpy config set dependencies.elrond_proxy_go.tag master
erdpy testnet prerequisites
erdpy testnet config
erdpy testnet start
```
Note: There was an issue when the `elrond_proxy_go` version was not specified.

claudiu725 (approved):

andrei-marinica (dismissed):

PR #144, opened 2022-05-29
Show password prompt if no passfile specificied

lcswillems:
Closes https://github.com/ElrondNetwork/elrond-sdk-erdpy/issues/66

Make --passfile argument optional. When a keyfile is specified without a corresponding passfile, a prompt will be shown asking for keyfile's password. Developers won't have to save the password on disk anymore.

Possible breaking change when using erdpy as a library: constructor of Account has been adjusted to take as input the keyfile's password directly, instead of the path of file containing the password.

### DIFF
--- a/erdpy/accounts.py
+++ b/erdpy/accounts.py
@@ -11,7 +11,7 @@ from erdpy.ledger.ledger_app_handler import SIGN_USING_HASH_VERSION
 from erdpy.ledger.ledger_functions import do_get_ledger_address, do_sign_transaction_with_ledger, do_get_ledger_version, \
     TX_HASH_SIGN_VERSION, TX_HASH_SIGN_OPTIONS
 from erdpy.wallet import bech32, pem
-from erdpy.wallet.keyfile import get_password, load_from_key_file
+from erdpy.wallet.keyfile import load_from_key_file
 
 logger = logging.getLogger("accounts")
 
@@ -22,7 +22,7 @@ class Account(IAccount):
                  pem_file: Optional[str] = None,
                  pem_index: int = 0,
                  key_file: str = "",
-                 pass_file: str = "",
+                 password: str = "",
                  ledger: bool = False):
         self.address = Address(address)
         self.pem_file = pem_file
@@ -34,8 +34,7 @@ class Account(IAccount):
             secret_key, pubkey = pem.parse(Path(self.pem_file), self.pem_index)
             self.secret_key = secret_key.hex()
             self.address = Address(pubkey)
-        elif key_file and pass_file:
-            password = get_password(pass_file)
+        elif key_file and password:
             address_from_key_file, secret_key = load_from_key_file(key_file, password)
             self.secret_key = secret_key.hex()
             self.address = Address(address_from_key_file)
--- a/erdpy/cli_contracts.py
+++ b/erdpy/cli_contracts.py
@@ -7,6 +7,7 @@ from pathlib import Path
 from erdpy import cli_shared, errors, projects, utils
 from erdpy.accounts import Account, Address, LedgerAccount
 from erdpy.cli_output import CLIOutputBuilder
+from erdpy.cli_password import load_password
 from erdpy.contracts import CodeMetadata, SmartContract
 from erdpy.projects import load_project
 from erdpy.projects.core import get_project_paths_recursively
@@ -284,8 +285,9 @@ def _prepare_sender(args: Any) -> Account:
         sender = LedgerAccount(account_index=args.ledger_account_index, address_index=args.ledger_address_index)
     elif args.pem:
         sender = Account(pem_file=args.pem, pem_index=args.pem_index)
-    elif args.keyfile and args.passfile:
-        sender = Account(key_file=args.keyfile, pass_file=args.passfile)
+    elif args.keyfile:
+        password = load_password(args)
+        sender = Account(key_file=args.keyfile, password=password)
     else:
         raise errors.NoWalletProvided()
 
--- a/erdpy/cli_shared.py
+++ b/erdpy/cli_shared.py
@@ -7,6 +7,7 @@ from typing import Any, List, Text, cast
 from erdpy import config, errors, scope, utils
 from erdpy.accounts import Account
 from erdpy.cli_output import CLIOutputBuilder
+from erdpy.cli_password import load_password
 from erdpy.ledger.ledger_functions import do_get_ledger_address
 from erdpy.proxy.core import ElrondProxy
 from erdpy.simulation import Simulator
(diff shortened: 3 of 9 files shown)


erdpy/cli_shared.py
```diff
@@ -80,7 +80,7 @@ def add_wallet_args(args: List[str], sub: Any):
     sub.add_argument("--pem", required=check_if_sign_method_required(args, "--pem"), help="🔑 the PEM file, if keyfile not provided")
     sub.add_argument("--pem-index", default=0, help="🔑 the index in the PEM file (default: %(default)s)")
     sub.add_argument("--keyfile", required=check_if_sign_method_required(args, "--keyfile"), help="🔑 a JSON keyfile, if PEM not provided")
-    sub.add_argument("--passfile", required=(utils.is_arg_present(args, "--keyfile")), help="🔑 a file containing keyfile's password, if keyfile provided")
+    sub.add_argument("--passfile", help="🔑 a file containing keyfile's password, if keyfile provided")
```
andreibancioiu, line 83: :+1:

erdpy/cli_contracts.py
```diff
@@ -284,7 +284,7 @@ def _prepare_sender(args: Any) -> Account:
         sender = LedgerAccount(account_index=args.ledger_account_index, address_index=args.ledger_address_index)
     elif args.pem:
         sender = Account(pem_file=args.pem, pem_index=args.pem_index)
-    elif args.keyfile and args.passfile:
+    elif args.keyfile:
```
andreibancioiu, line 287: This change should be applied in several other places, as well (since the logic is currently duplicated in erdpy, unfortunately):
 - `cli_contracts.py` (updated by you)
 - `cli_shared.py`
 - `transactions.py`
 - `staking_provider.py`
 - `validators/core.py`

andreibancioiu: However, how about not passing a `pass_file` at all (whether optional or not) - and pass the `password`, as a string. And possibly add a function in `cli_shared.py`, such as:

```
def load_password(args):
    (loads from file if provided)
    (loads from stdin otherwise)
    returns the password, as a string
```

Then call it as follows:

```
else if args.keyfile is present:
    password = cli_shared.load_password(args)
    account = Account(keyfile = ..., password = password)
```

How do you feel about this change?

erdpy/accounts.py
```diff
@@ -34,8 +35,11 @@ def __init__(self,
             secret_key, pubkey = pem.parse(Path(self.pem_file), self.pem_index)
             self.secret_key = secret_key.hex()
             self.address = Address(pubkey)
-        elif key_file and pass_file:
-            password = get_password(pass_file)
+        elif key_file:
```
andreibancioiu, line 38: Ideally, we should not open files or ask for input in the `Account` class. Indeed, currently we read and parse some files here, directly in the constructor (unfortunately).

However, how do you feel about removing the `pass_file` argument of the constructor, and pass a `password: optional string`, instead? (breaking change, but an improvement)

(see comment below)

erdpy/accounts.py
```diff
@@ -22,7 +23,7 @@ def __init__(self,
                  pem_file: Optional[str] = None,
                  pem_index: int = 0,
                  key_file: str = "",
-                 pass_file: str = "",
+                 pass_file: Optional[str] = None,
```
andreibancioiu, line 26: Perhaps pass the actual password, loaded in advance by the CLI components? (see the other comments)

erdpy/accounts.py
```diff
@@ -1,6 +1,7 @@
 import logging
 from pathlib import Path
 from typing import Any, Optional
+from getpass import getpass
```
andreibancioiu, line 4: Nice :+1: (only now I learn that this module is in the standard library - it seems we also use it in `cli_wallet.py`).

andreibancioiu (commented):
Added some suggestions / loading the password in advance, not in `Account` (these changes would ultimately be breaking - although there are some workarounds to avoid breaking changes).

lcswillems:
I have followed all the feedback. Let me know if everything is ok!

erdpy/cli_shared.py
```diff
@@ -80,7 +81,8 @@ def add_wallet_args(args: List[str], sub: Any):
...
+    sub.add_argument("--password", help="🔑 the keyfile's password, if keyfile provided")
```
andreibancioiu, line 85: Not sure we want this (or perhaps it can be safely passed as ENV variable). Will double check with another reviewer.

erdpy/CLI.md
```diff
@@ -203,6 +203,7 @@ optional arguments:
   --pem-index PEM_INDEX                        🔑 the index in the PEM file (default: 0)
   --keyfile KEYFILE                            🔑 a JSON keyfile, if PEM not provided
   --passfile PASSFILE                          🔑 a file containing keyfile's password, if keyfile provided
+  --password PASSWORD                          🔑 the keyfile's password, if keyfile provided
```
andreibancioiu, line 206: If we don't want this parameter (see comments below), then we should revert this file as well.

andreibancioiu (commented):
Very nice refactoring :rocket: 

Regarding the `--password` CLI parameter, I will double check with another reviewer then approve the PR.

Thank you :pray: 

Also, do you think you can update the PR description to include the text `possible breaking change when using erdpy as a library: constructor of Account has been adjusted [...]`.

andreibancioiu:
Also, @lcswillems, I think we can extract `load_password(args)` out of `cli_shared.py`, since it causes a cyclic import:

https://github.com/ElrondNetwork/elrond-sdk-erdpy/runs/6708800344?check_suite_focus=true#step:7:132

Perhaps `password.py` (or something similar) can be its new home.

erdpy/cli_shared.py
```diff
@@ -186,3 +189,11 @@ def check_if_sign_method_required(args: List[str], checked_method: str) -> bool:
             return False
 
     return True
+
+def load_password(args):
+    if args.passfile:
+        with open(args.passfile) as pass_file:
+            return pass_file.read().strip()
+    if args.password:
```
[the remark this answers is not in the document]
ccorcoveanu: I would **only** go with the prompt. This is a long awaited feature, thanks [lcswillems](https://github.com/lcswillems)!

erdpy/CLI.md
lcswillems: Removed

erdpy/cli_shared.py
lcswillems: Removed too

erdpy/cli_shared.py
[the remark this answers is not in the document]
lcswillems: I removed `--password` argument

lcswillems:
Thank you @andreibancioiu for your review!

I removed `--password` argument and I moved the `load_function` to `cli_password` file as it is a method related to CLI arguments but let me know if you want me to change the file name.

andreibancioiu (approved):

erdpy/cli_shared.py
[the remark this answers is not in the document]
andreibancioiu: Thank you :pray: (sorry for being ambiguous before, I was referring only to the `Account` constructor argument).

:heavy_check_mark:

ccorcoveanu (approved):

PR #146, opened 2022-06-12
Example erdpy functions: Issuance of non fungible token

tommyhutcheson:
Example erdpy functions: Issuance of non fungible token, get token identifier and set special role for token issued.

I want to contribute some python functions for creating NFTs, this PR issues a new token issuance but I have not got the set role nft function correct here as I need to first get the newly created token identifier back, can someone review my PR and suggest changes.

Thank you


danidrasovean:
Will be handled by https://github.com/multiversx/mx-sdk-py-cli/issues/325.

PR #151, opened 2022-09-09
Add support for reproducible builds

andreibancioiu:


andreibancioiu:
Dropped experiment.

Recommended approach (as of October 2022): https://docs.elrond.com/developers/reproducible-contract-builds.

PR #164, opened 2022-12-16
Reproducible builds endpoint for erdpy cli

popenta:
Added a new endpoint called `reproducible-build` for erdpy cli.
The purpose of this endpoint is to reproduce a previously built Smart Contract using a Docker image.
The command can be used as easy as: **erdpy contract reproducible-build [project-path] --contract=[contract-name] 
--docker-image=[docker-image]**.
Build with docker implementation: https://github.com/ElrondNetwork/elrond-sdk-images-build-contract-rust/blob/main/build_with_docker.py
Also added a test.

### DIFF
--- a/erdpy/cli_contracts.py
+++ b/erdpy/cli_contracts.py
@@ -12,6 +12,7 @@ from erdpy.projects import load_project
 from erdpy.projects.core import get_project_paths_recursively
 from erdpy.proxy.core import ElrondProxy
 from erdpy.transactions import Transaction
+from erdpy.docker import is_docker_installed, run_docker
 
 logger = logging.getLogger("cli.contracts")
 
@@ -120,6 +121,15 @@ def setup_parser(args: List[str], subparsers: Any) -> Any:
     _add_function_arg(sub)
     _add_arguments_arg(sub)
     sub.set_defaults(func=query)
+
+    sub = cli_shared.add_command_subparser(subparsers, "contract", "reproducible-build",
+                                            "Build a Smart Contract and get the same output as a previously built Smart Contract")
+    _add_project_arg(sub)
+    _add_recursive_arg(sub)
+    _add_build_options_args(sub)
+    sub.add_argument("--docker-image-tag", required=True, type=str,
+                        help="the docker image tag used to build the contract")
+    sub.set_defaults(func=reproduce_build)
 
     parser.epilog = cli_shared.build_group_epilog(subparsers)
     return subparsers
@@ -359,3 +369,19 @@ def _send_or_simulate(tx: Transaction, contract: SmartContract, args: Any):
     output_builder = cli_shared.send_or_simulate(tx, args, dump_output=False)
     output_builder.set_contract_address(contract.address)
     utils.dump_out_json(output_builder.build(), outfile=args.outfile)
+
+
+def reproduce_build(args: Any):
+    project_path = get_project_paths(args)[0]
+    docker_image = args.docker_image_tag
+    output_path = project_path / "output-docker"
+    contract_path = ""
+
+    options = _prepare_build_options(args)
+    # cargo_target_dir = Path(options.get("cargo-target-dir", ""))
+    no_wasm_opt = options.get("no-wasm-opt", True)
+
+    if is_docker_installed():
+        return_code = run_docker(docker_image, project_path, contract_path, output_path, None, no_wasm_opt)
+
+    return return_code
--- a/erdpy/cli_contracts.py
+++ b/erdpy/cli_contracts.py
@@ -127,8 +127,9 @@ def setup_parser(args: List[str], subparsers: Any) -> Any:
     _add_project_arg(sub)
     _add_recursive_arg(sub)
     _add_build_options_args(sub)
-    sub.add_argument("--docker-image-tag", required=True, type=str,
+    sub.add_argument("--docker-image", required=True, type=str,
                         help="the docker image tag used to build the contract")
+    sub.add_argument("--contract", type=str, help="relative path of the contract in the project")
     sub.set_defaults(func=reproduce_build)
 
     parser.epilog = cli_shared.build_group_epilog(subparsers)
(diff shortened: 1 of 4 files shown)


lcswillems:
Very great PR @popenta ! Currently, I need to do 20 minutes calls with people who want to do reproducible builds, but with this PR, I will only need to send them a short text message. So this PR is going to improve a lot the experience of building smart contracts in a reproducible way.

I see we always have to provide a `[project-path]`. Most of the time, I think people will execute the command at the root of the project. So maybe this should be made optional and simply a `--project` flag?

popenta:
> 

Thank you!
If the user executes the command at the root of the project the [`project-path`] arg should not be provided as it takes the current directory as default. I've discussed with @andreibancioiu to make it an optional argument but we've decided to keep it standard as a positional argument across erdpy-cli. Maybe we'll change it in the future.

lcswillems:
That's great if `project-path` is not mandatory. And it is better to keep consistency accross the commands indeed. From a developer experience point of view, it looks go to me.

erdpy/errors.py
```diff
@@ -188,3 +188,8 @@ def __init__(self, message: str):
 class LedgerError(KnownError):
     def __init__(self, message: str):
         super().__init__("Ledger error: " + message)
+
+
+class DockerMissingError(KnownError):
+    def __init__(self):
+        super().__init__("Docker is not installed! Please visit https://docs.docker.com/get-docker/ to install docker.")
```
andreibancioiu, line 195: :+1:

erdpy/cli_contracts.py
```diff
@@ -121,6 +122,16 @@ def setup_parser(args: List[str], subparsers: Any) -> Any:
     _add_arguments_arg(sub)
     sub.set_defaults(func=query)
 
+    sub = cli_shared.add_command_subparser(subparsers, "contract", "reproducible-build",
+                                            "Build a Smart Contract and get the same output as a previously built Smart Contract")
+    _add_project_arg(sub)
+    _add_recursive_arg(sub)
```
andreibancioiu, line 128: Here, I think we do not need the `_add_recursive_arg`. Remove if redundant.

erdpy/cli_contracts.py
```diff
@@ -359,3 +370,28 @@ def _send_or_simulate(tx: Transaction, contract: SmartContract, args: Any):
     output_builder = cli_shared.send_or_simulate(tx, args, dump_output=False)
     output_builder.set_contract_address(contract.address)
     utils.dump_out_json(output_builder.build(), outfile=args.outfile)
+
+
+def reproduce_build(args: Any):
```
andreibancioiu, line 375: Optionally, can be renamed to `do_reproducible_build()`.

erdpy/cli_contracts.py
```diff
@@ -359,3 +370,28 @@ def _send_or_simulate(tx: Transaction, contract: SmartContract, args: Any):
...
+    project_paths = get_project_paths(args)
+
+    if len(project_paths) == 1:
+        project_path = project_paths[0].expanduser().resolve()
+    else:
+        raise NotImplementedError()
```
andreibancioiu, line 381: I think, in this case, a simplification would be to not use `_add_recursive_arg`, nor `get_project_paths()`, and just use `args.project_path`, instead.

Can also be kept as it is, for the moment (`_add_recursive_arg` + `get_project_paths`).

erdpy/docker.py
```diff
@@ -0,0 +1,57 @@
...
+import subprocess
+from pathlib import Path
+from typing import List
+from erdpy.errors import DockerMissingError
+
+
+logger = logging.getLogger("build-with-docker")
+
+
+def is_docker_installed():
```
andreibancioiu, line 12: Rename to `assert_docker_is_installed()`? Or make it to return a bool, instead of throwing an error, and throw it above, where called?

(usually, `is_...` functions should return boolean values).

erdpy/docker.py
```diff
@@ -0,0 +1,57 @@
...
+    try:
+        result = subprocess.run(["docker", "--version"], capture_output=True, text=True)
```
andreibancioiu, line 14: Actually, I think this will raise an exception if docker is not installed.

Can you make a (manual) test, say, using `["dummy-test-docker", "--version"]`?

Also, `subprocess.check_output()` can be used, perhaps?

erdpy/docker.py
```diff
@@ -0,0 +1,57 @@
...
+        output = result.stdout
+    
+        if "Docker version" in output:
+            return True
+        else:
+            raise DockerMissingError()
+    except:
+        logger.error("Something went wrong when checking if docker is installed!")
```
andreibancioiu, line 22: I am not sure, isn't this dead code?

erdpy/docker.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def run_docker(image: str, project_path: Path, contract: str, output_path: Path, no_wasm_opt: bool):
+    docker_mount_args: List[str] = ["--volume", f"{output_path}:/output"]
+
+    if project_path:
+        docker_mount_args.extend(["--volume", f"{project_path}:/project"])
+
+
```
andreibancioiu, line 31: Formatting / one vs. two blank lines.

erdpy/docker.py
```diff
@@ -0,0 +1,57 @@
...
+    docker_args = ["docker", "run"]
+
+    docker_args += ["--interactive"]
+    docker_args += ["--tty"]
+    docker_args += docker_mount_args
+    docker_args += ["--user", f"{str(os.getuid())}:{str(os.getgid())}"]
```
andreibancioiu, line 37: In the PR description, add a link to https://github.com/ElrondNetwork/elrond-sdk-images-build-contract-rust/blob/main/build_with_docker.py.

erdpy/cli_contracts.py
popenta: Renamed it!

erdpy/cli_contracts.py
popenta: Deleted it!

erdpy/cli_contracts.py
popenta: Modified to only use `args.project`

erdpy/docker.py
popenta: Changed `is_docker_installed` to return a boolean and moved `raise DockerMissingEror` on else statement.

erdpy/docker.py
popenta: Actually, it doesn't raise an error, it says "dummy-test-docker: command not found". I expect `subprocess.check_output` to have the same behaviour, but it's recomended to use `subprocess.run`

erdpy/docker.py
popenta: Because now the function does not raise an error anymore, I expect that if for some reason something goes wrong with the `subprocess.run` the except statement gets executed.

erdpy/docker.py
popenta: Two blank lines were left on accident :)

andreibancioiu (approved):

claudiu725 (approved):

PR #184, opened 2023-01-13
Package renaming

popenta:
Fixes for the migration. 
The word `erdpy` it's still used in some places in `erdpy-up.py`, but that is because `mxpy` will not be installed like this, it will use a new installer (cc: @andreibancioiu ).


claudiu725 (approved):

multiversx_sdk_cli/tests/test_cli_contracts.sh
```diff
@@ -130,7 +130,7 @@ testVerifyContract(){
 
     command_response=$(${CLI} contract verify erd1qqqqqqqqqqqqqpgquzmh78klkqwt0p4rjys0qtp3la07gz4d396qn50nnm \
                         --verifier-url=http://localhost:7777 --packaged-src=testdata/dummy.json \
-                        --pem=testdata/walletKey.pem --docker-image=elrondnetwork/build-contract-rust:v4.0.0)
+                        --pem=testdata/walletKey.pem --docker-image=multiversx/sdk-rust-contract-builder:v4.0.0)
```
andreibancioiu, line 133: :+1: (we have some newer versions, but it's all right like this for the moment)

andreibancioiu (approved):

PR #187, opened 2023-01-17
Remove mcl signer dependency

popenta:
Removed the mcl signer dependency by using the `ValidatorSigner` from the `mx-py-wallet` package.


multiversx_sdk_cli/accounts.py
```diff
@@ -141,6 +141,9 @@ def __repr__(self):
     @classmethod
     def zero(cls) -> 'Address':
         return Address("0" * 64)
+    
+    def serialize_for_signing(self) -> bytes:
```
andreibancioiu, line 145: Oh, I see why :)

multiversx_sdk_cli/validators/core.py
```diff
@@ -56,7 +56,13 @@ def prepare_transaction_data_for_stake(node_operator_address: Address, validator
         validator_pem = validator_pem if validator_pem.is_absolute() else validators_file_path / validator_pem
         
         secret_key_bytes, bls_key = parse_validator_pem(Path(validator_pem))
-        signed_message = sign_message_with_bls_key(node_operator_address.hex(), secret_key_bytes.decode('ascii'))
+        secret_key_str = secret_key_bytes.decode()
+
+        validator_secret_key = ValidatorSecretKey.from_string(secret_key_str)
```
andreibancioiu, line 61: :+1:

multiversx_sdk_cli/interfaces.py
```diff
@@ -39,9 +39,12 @@ def get_data(self) -> str:
         return ""
 
 
-class IAccount:
+class IAccount(Protocol):
     def sign_transaction(self, transaction: ITransaction) -> str:
-        return ""
+        ...
+    
+    def serialize_for_signing(self) -> bytes:
```
andreibancioiu, line 46: All right for now. An alternative could have been to create a `Message` such as `message = Message(account.pubkey)` and pass that to the `ValidatorSigner`. `Message` is a class from multiversx-sdk-core.

multiversx_sdk_cli/delegation/staking_provider.py
```diff
@@ -47,7 +49,13 @@ def prepare_args_for_add_nodes(args: Any):
         validator_pem = validator_pem if validator_pem.is_absolute() else validators_file_path / validator_pem
 
         secret_key_bytes, bls_key = parse_validator_pem(validator_pem)
-        signed_message = sign_message_with_bls_key(account.address.pubkey().hex(), secret_key_bytes.decode('ascii'))
+        secret_key_str = secret_key_bytes.decode()
+
+        validator_secret_key = ValidatorSecretKey.from_string(secret_key_str)
```
andreibancioiu, line 54: An alternative (as written in the other comment) would be to use the `Message` class (which already implements `ISignable`).

andreibancioiu (approved):
:evergreen_tree:, but with a design change proposal (using `Message` instead of implementing `ISignable` in `Account`).

andreibancioiu (approved):

schimih (approved):

PR #201, opened 2023-01-25
mxpy contract verify fixes

popenta:
These are the fixes that needed to be done after a discussion with @andreibancioiu , @tanghel and @florincoz .
I've tested the fixes on my local machine, but feel free to test them on your own machine.


multiversx_sdk_cli/cli_contracts.py
```diff
@@ -131,14 +131,8 @@ def setup_parser(args: List[str], subparsers: Any) -> Any:
...
 
-    group = sub.add_mutually_exclusive_group(required=True)
-    group.add_argument(
-        "--project",
-        default=os.getcwd(),
-        help="🗀 the project directory (default: current directory)",
-    )
-    group.add_argument(
-        "--packaged-src", help="JSON file containing the source code of the contract"
+    sub.add_argument(
```
andreibancioiu, line 134: :+1: we require `packaged-src` now.

multiversx_sdk_cli/cli_contracts.py
```diff
@@ -147,7 +141,7 @@ def setup_parser(args: List[str], subparsers: Any) -> Any:
         required=True,
         help="the url of the service that validates the contract",
     )
-    sub.add_argument("--docker-image", required=True, help="the docker image used for the build (i.e. multiversx/sdk-rust-contract-builder:v4.0.0)")
+    sub.add_argument("--docker-image", required=True, help="the docker image used for the build (i.e. multiversx/sdk-rust-contract-builder:next)")
```
andreibancioiu, line 144: Let's remove the default, and only decide on a default value one week from now?

andreibancioiu (approved):

andreibancioiu (approved):

claudiu725 (approved):

multiversx_sdk_cli/contract_verification.py
```diff
@@ -80,6 +77,38 @@ def trigger_contract_verification(
...
+    status = response.get("status", "")
+    if status:
+        logger.info(f"Task status: {status}")
+
+        if status == "error":
+            dump_out_json(response)
+        else:
+            dump_out_json(response)
+    else:
+        task_id = response.get("taskId", "")
```
tanghel, line 93: an error can be also of other causes, so taskId should be checked here, and if it does not exist, dump the error in a standard way

popenta: will do!

PR #206, opened 2023-01-28
Remove local testnet scope. It caused bugs in the past, without being very useful

andreibancioiu:
Remove `local testnet scope`. It caused bugs in the past, without being very useful.


andreibancioiu:
Should not be approved as it is. I have to undo fields removal from `config.py`, since `config.add_config_args` and `config.def determine_final_args` depend on them.

miiu96 (approved):

bogdan-rosianu (approved):

claudiu725 (approved):

PR #212, opened 2023-01-31
Fix "wallet new --pem"

andreibancioiu:
```
mxpy wallet new --pem
```
Resulted in an error (when creating the label for the PEM file).

Fixed regression, added CLI test.


camilbancioiu (approved):

multiversx_sdk_cli/cli_wallet.py
```diff
@@ -95,7 +95,7 @@ def new_wallet(args: Any):
 
     if args.pem:
         pem_file = prepare_file(args.output_path, ".pem")
-        address = Address(pubkey)
+        address = Address(pubkey.buffer)
```
mariusmihaic, line 98: Can pubkey be `nil` so you would need to check before calling `.buffer` ?

mariusmihaic (commented):
Good job

multiversx_sdk_cli/cli_wallet.py
andreibancioiu: It cannot, it's generated here:

```
pubkey = secret_key.generate_public_key()
```

mariusmihaic (approved):

claudiu725 (approved):

PR #217, opened 2023-02-06
Set default platform by default + flag to skip setting it

lcswillems:
A developer on Mac M1 who wants to do a reproducible build has to first set the `DOCKER_DEFAULT_PLATFORM` environment variable to "linux/amd64". If he/she doesn't know or forgets, the build might not be reproducible.

Having to think about adding this flag is friction for reproducible builds. Hence `mxpy` would set it by default. As setting this flag can significantly slow the build process on Mac M1, we let the option to skip setting it with the flag `--no-default-platform`.

Docker documentation about `--platform` and `DOCKER_DEFAULT_PLATFORM`:

> DOCKER_DEFAULT_PLATFORM | Default platform for commands that take the --platform flag.

https://docs.docker.com/engine/reference/commandline/cli/#environment-variables

### DIFF
--- a/multiversx_sdk_cli/docker.py
+++ b/multiversx_sdk_cli/docker.py
@@ -31,6 +31,7 @@ def run_docker(
         no_wasm_opt: bool,
         docker_interactive: bool,
         docker_tty: bool,
+        no_default_platform: bool
 ):
     docker_mount_args: List[str] = ["--volume", f"{output_path}:/output"]
 
@@ -47,6 +48,9 @@ def run_docker(
     docker_args += docker_mount_args
     docker_args += ["--user", f"{str(os.getuid())}:{str(os.getgid())}"]
     docker_args += ["--rm", image]
+
+    if not(no_default_platform):
+        docker_args += ["--platform", "linux/amd64"]
 
     entrypoint_args: List[str] = []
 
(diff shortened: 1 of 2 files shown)


andreibancioiu (dismissed):

multiversx_sdk_cli/docker.py
```diff
@@ -48,6 +49,9 @@ def run_docker(
     docker_args += ["--user", f"{str(os.getuid())}:{str(os.getgid())}"]
     docker_args += ["--rm", image]
 
+    if not(no_default_platform):
+        docker_args += ["--platform", "linux/amd64"]
```
andreibancioiu, line 53: I am unaware if this parameter influences Docker at the same level as `DOCKER_DEFAULT_PLATFORM`.

lcswillems: https://docs.docker.com/engine/reference/commandline/cli/#environment-variables
> DOCKER_DEFAULT_PLATFORM | Default platform for commands that take the --platform flag.

lcswillems: But I have not tested on my computer in order to be 100% sure

lcswillems: Also there is this one: https://stackoverflow.com/questions/65612411/forcing-docker-to-use-linux-amd64-platform-by-default-on-macos

bogdan-rosianu (dismissed):

bogdan-rosianu (approved):

PR #218, opened 2023-02-08
Extra logging for contract verify

andreibancioiu:
Add extra logging for `contract verify`.

Also fix configuration for general logging.


multiversx_sdk_cli/cli.py
```diff
@@ -45,9 +45,9 @@ def _do_main(cli_args: List[str]):
     args = parser.parse_args(argv_with_config_args)
 
     if args.verbose:
-        logging.basicConfig(level=logging.DEBUG)
+        logging.basicConfig(level=logging.DEBUG, force=True)
```
andreibancioiu, line 48: Without `force = True`, already created logger objects (all of them, that is) won't have the new config.

bogdan-rosianu (approved):

bogdan-rosianu (approved):

popenta (approved):

PR #225, opened 2023-02-13
mxpy-up improvements. mxpy not auto-registered to PATH anymore. Partial support for Windows.

andreibancioiu:
## `PATH` not altered by default anymore

In `mxpy-up`, we do not alter the `PATH` variable by default anymore; that is, we do not alter `.profile`, `.zshrc` or `.bash_profile` automatically anymore. The user has to explicitly demand this (during the installation script).

## Experimental, partial support for Windows

**BASH** shell required, though (at least for installation).

Users can now use the `mxpy-up` facility to install `mxpy` in Windows, outside of WSL. However, Windows support is **experimental** and **partial** at the moment. Contract builds aren't available yet (due to a few issues related to the Rust installation) - they will be in a future version, however. Transaction creation / broadcasting and wallet utilities are available. 

## Other changes
 - Fixed `mxpy` shortcut to use `mxpy-venv`. Prior v6, the virtual environment wasn't correctly activated when invoking the `mxpy` command.


multiversx_sdk_cli/dependencies/modules.py
```diff
@@ -253,18 +253,36 @@ def get_latest_release(self) -> str:
...
-        args = [rustup_path, "--verbose", "--default-toolchain", toolchain, "--profile",
+        args = [str(installer_path), "--verbose", "--default-toolchain", toolchain, "--profile",
                 "minimal", "--target", "wasm32-unknown-unknown", "--no-modify-path", "-y"]
         myprocess.run_process(args, env=self.get_env_for_install())
 
+    def _get_installer_url(self) -> str:
+        platform = workstation.get_platform()
+
+        if platform == "windows":
+            return "https://win.rustup.rs"
```
andreibancioiu, line 275: Added, but doesn't work out of the box (downloads & certificates-related issues).

mxpy-up.py
```diff
@@ -16,38 +17,34 @@
 
 def main():
     parser = ArgumentParser()
-    parser.add_argument("--modify-path", dest="modify_path", action="store_true", help="whether to modify $PATH (in profile file)")
-    parser.add_argument("--no-modify-path", dest="modify_path", action="store_false", help="whether to modify $PATH (in profile file)")
+    parser.add_argument("--modify-path", dest="modify_path", action="store_true", help="(deprecated, not used)")
```
andreibancioiu, line 20: Kept for one more release, to ensure a smooth transition to mxpy v6 in IDE, as well - the VSCode extension was setting `--no-modify-path`.

lcswillems (dismissed):

popenta (approved):

mxpy-up.py
andreibancioiu: Not kept anymore. We'll prepare a hotfix for the IDE when mxpy 6 is released.

mxpy-up.py
```diff
@@ -321,13 +298,154 @@ def run_post_install_checks():
...
 
 
+def add_shortcut_to_system_path():
+    interactive = True
+    operating_system = get_operating_system()
+
+    if operating_system == "windows":
+        print(f"""
+###############################################################################
+On Windows, for the "mxpy" command shortcut to be available, you need to add the the directory "{sdk_path}" to the system PATH.
```
lcswillems, line 308: "the the" -> "the"

mxpy-up.py
```diff
@@ -321,13 +298,154 @@ def run_post_install_checks():
...
+In order to manually extend the PATH variable, add the following line to your shell profile file upon installation:
+
+    export PATH="${{HOME}}/multiversx-sdk:${{PATH}}"
+
+Your shell profile files:
+{profile_files_formatted}
+
+Upon editing the shell profile file, you may have to RESTART THE USER SESSION for the changes to take effect.
+
+Please pick an option:
```
lcswillems, line 397: I don't know if it is clear that we need to input "a" or "b" in order to pick an option.

mxpy-up.py
```diff
@@ -321,13 +298,154 @@ def run_post_install_checks():
...
+    for profile_file in profile_files:
+        with open(profile_file, "a") as file:
+            file.write(f'\n{new_export_directive}\n')
+
+    print(f"""
+###############################################################################
+The following profile files have been updated:
+{profile_files_formatted}
+
+Please RESTART THE USER SESSION.
```
lcswillems, line 418: By the way, why adding to the `.profile` file and not the `.bashrc` one? If we add to the latter, we don't need to restart the user session. I imagine it is because `.profile` exists on more OSes

lcswillems (approved):

mxpy-up.py
```diff
@@ -226,130 +233,239 @@ def install_mxpy(exact_version: str, from_branch: str):
...
-    profile_file = get_profile_file()
-    profile_info_content = profile_file.read_text()
+    profile_files = get_profile_files()
+
+    if not profile_files:
+        print(f"""
+###############################################################################
+No shell profile files have been found.
+
+The "mxpy" command shortcut will not be available until you add the the directory "{sdk_path}" to the system PATH.
```
popenta, line 331: `the the` again

mxpy-up.py
```diff
@@ -226,130 +233,239 @@ def install_mxpy(exact_version: str, from_branch: str):
...
         # Note: in some (rare) cases, here we'll have false positives (e.g. if the export directive is commented out).
-        logger.info(f"multiversx-sdk path ({sdk_path}) is already configured in shell profile.")
-        return
+        print(f"""
+###############################################################################
+It seems that the path "~/multiversx-sdk" is already configured in shell profile.
 
-    logger.info(f"Configuring multiversx-sdk path [{sdk_path}] in shell profile...")
-    logger.info(f"[{profile_file}] is being modified...")
+To confirm this fact, check the shell profile (now or after the installer script ends).
```
popenta, line 368: Maybe remove the word `fact`? Something like `To confirm this, please check ...` sounds a little better in my opinion.

mxpy-up.py
andreibancioiu: Fixed.

popenta (dismissed):

mxpy-up.py
andreibancioiu: Fixed, added an extra line.

mxpy-up.py
andreibancioiu: Good point, handling `.bashrc` as well now.

Indeed, more experienced users will know that only the shell needs to be restarted in this case. But leaving the message "restart the user session" or even "restart the computer" could act as a helpful "fallback".

mxpy-up.py
andreibancioiu: Fixed.

mxpy-up.py
andreibancioiu: Fixed.

popenta (approved):

lcswillems (approved):

PR #234, opened 2023-02-20
Do not alter shell profile files anymore

andreibancioiu:
Instead, users are instructed on how to add `~/multiversx-sdk` to the `PATH` environment variable.

**Reason:** less side-effects on the system configuration should lead to less installation issues. Users are responsible with altering the system path, if desired.


popenta (approved):

popenta:
Maybe also increase the version and release a new version? Since this is a hotfix.

schimih (approved):

PR #235, opened 2023-03-08
Allow keystore files with mnemonic (new format). Implement "wallet convert"

andreibancioiu:
- Re-implement command `wallet new` (non-breaking change)
 - Add a new command `wallet convert` to convert between different types of wallets
 - Add support for the new type of keystore files (that hold an encrypted mnemonic)


multiversx_sdk_cli/cli_wallet.py
```diff
@@ -89,38 +111,155 @@ def setup_parser(args: List[str], subparsers: Any) -> Any:
...
+    if outfile:
+        outfile.write_text(output_text)
+    else:
+        print("Output:")
+        print()
+        print(output_text)
+
+
+def _load_wallet(input_text: str, in_format: str, address_index: int) -> Tuple[Optional[Mnemonic], Optional[UserSecretKey]]:
+    if in_format == WALLET_FORMAT_RAW_MNEMONIC:
```
bogdan-rosianu, line 194: not sure about the required python version, but it seems like [>=3.10 supports switches](https://www.freecodecamp.org/news/python-switch-statement-switch-case-example/)

multiversx_sdk_cli/cli_wallet.py
```diff
@@ -28,41 +41,50 @@ def setup_parser(args: List[str], subparsers: Any) -> Any:
...
+    sub = cli_shared.add_command_subparser(
+        subparsers,
+        "wallet",
+        "convert",
+        "Convert a wallet from one format to another"
+    )
+
+    sub.add_argument("--infile", help="path to the input file")
+    sub.add_argument("--outfile", help="path to the output file")
+    sub.add_argument("--in-format", required=True, choices=WALLET_FORMATS, help="the format of the input file")
```
bogdan-rosianu, line 72: can all the possible options be listed in the help? are they by default?

multiversx_sdk_cli/cli_wallet.py
```diff
@@ -89,38 +111,155 @@ def setup_parser(args: List[str], subparsers: Any) -> Any:
...
+    address_hrp = args.address_hrp
 
-        logger.info(f"Json wallet generated: {json_file}")
+    if outfile and outfile.exists():
+        raise KnownError(f"File already exists, will not overwrite: {outfile}")
+
+    if infile:
+        input_text = infile.read_text()
+    else:
+        print(f"Press 'Ctrl-D' (Linux / MacOS) or 'Ctrl-Z' (Windows) when done.")
```
bogdan-rosianu, line 179: as discussed, might prepend a `Insert text` to this message

multiversx_sdk_cli/cli_wallet.py
```diff
@@ -89,38 +111,155 @@ def setup_parser(args: List[str], subparsers: Any) -> Any:
...
+        password = getpass.getpass("Enter the password for the input keystore:")
+        keyfile = json.loads(input_text)
+        secret_key = UserWallet.decrypt_secret_key(keyfile, password)
+        return None, secret_key
+
+    if in_format == WALLET_FORMAT_PEM:
+        secret_key = UserPEM.from_text(input_text, address_index).secret_key
+        return None, secret_key
+
+    raise KnownError(f"Cannot load wallet, unknown input format: {in_format}")
```
bogdan-rosianu, line 215: might print all the available options, so the output would look like `Invalid wallet input format <{in_format}>. Make sure you use one of these: [raw-mnemonic, keystore-mnemonic, keystore-secret-key, pem]`

CLI.md
```diff
@@ -958,15 +965,39 @@ optional arguments:
...
+$ mxpy wallet convert --help
+usage: mxpy wallet convert [-h] ...
+
+Convert a wallet from one format to another
+
+optional arguments:
+  -h, --help                                      show this help message and exit
+  --infile INFILE                                 path to the input file
+  --outfile OUTFILE                               path to the output file
+  --in-format {raw-mnemonic,keystore-mnemonic,keystore-secret-key,pem}
```
bogdan-rosianu, line 992: nitpicking, but I think that the input format can be programmatically figured out (might be a new flag `--detect-in-format`)

multiversx_sdk_cli/cli_wallet.py
andreibancioiu: Currently, the constraint for mxpy is `>=3.8`. I think we should postpone this change. But very good point, indeed!

multiversx_sdk_cli/cli_wallet.py
andreibancioiu: They are listed in the help in the following format:

https://github.com/multiversx/mx-sdk-py-cli/blob/wallet-07/CLI.md#walletconvert

multiversx_sdk_cli/cli_wallet.py
andreibancioiu: Fixed.

multiversx_sdk_cli/cli_wallet.py
andreibancioiu: Added. Normally, this is unreachable (since the CLI arguments parser applies the constraints beforehand).

schimih (commented):
Good implementation :fire: . Only thing I would add are some tests with the new keystore format (the one with the kind field on both cases, secretKey and mnemonic).

bogdan-rosianu (dismissed):

bogdan-rosianu (approved):

schimih (approved):

PR #249, opened 2023-04-26
Config file for CLI args: --argsconfig

andreibancioiu:


andreibancioiu:
PR will be closed. CLI was refactored in the meantime. `--argsconfig` brought by this PR would be suboptimal (from a user's perspective). Will be brainstormed in the future (different design).

PR #266, opened 2023-05-29
Remove project argument from contract deploy and contract upgrade

popenta:
The `--project` argument from the `contract deploy` and `contract upgrade` commands has been removed.
It used to create confusion because when it was used, if the contract wasn't previously built it would build the contract first, then load the bytecode. If the contract was built, it wouldn't rebuild it, it would just load the existing bytecode.

### DIFF
--- a/multiversx_sdk_cli/cli_contracts.py
+++ b/multiversx_sdk_cli/cli_contracts.py
@@ -15,7 +15,6 @@ from multiversx_sdk_cli.contract_verification import \
 from multiversx_sdk_cli.contracts import CodeMetadata, SmartContract
 from multiversx_sdk_cli.docker import is_docker_installed, run_docker
 from multiversx_sdk_cli.errors import DockerMissingError
-from multiversx_sdk_cli.projects import load_project
 from multiversx_sdk_cli.projects.core import get_project_paths_recursively
 from multiversx_sdk_cli.transactions import Transaction
 from multiversx_sdk_cli.cosign_transaction import cosign_transaction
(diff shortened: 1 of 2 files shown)


danielailie (dismissed):

multiversx_sdk_cli/cli_shared.py
```diff
@@ -180,6 +180,11 @@ def check_options_for_guarded_tx(options: int):
         raise errors.BadUsage("Invalid guarded transaction's options. The second least significant bit must be set.")
 
 
+def check_bytecode_args(args: Any):
+    if not args.bytecode and len(args.bytecode):
```
andreibancioiu, line 184: We should declare it as `required` in `_add_bytecode_arg()`. Then, we will not need this validation anymore.

multiversx_sdk_cli/cli_contracts.py
```diff
@@ -191,12 +190,9 @@ def _add_recursive_arg(sub: Any):
...
 
 
-def _add_project_or_bytecode_arg(sub: Any):
-    group = sub.add_mutually_exclusive_group(required=True)
-    group.add_argument("--project", default=os.getcwd(),
-                       help="🗀 the project directory (default: current directory)")
-    group.add_argument("--bytecode", type=str,
-                       help="the file containing the WASM bytecode")
+def _add_bytecode_arg(sub: Any):
+    sub.add_argument("--bytecode", type=str,
```
andreibancioiu, line 194: Should be marked as `required`  here.

popenta: Set `required=True`

multiversx_sdk_cli/cli_shared.py
popenta: Deleted this function.

danielailie (approved):

PR #278, opened 2023-06-23
Replace transaction class with the one from sdk-core

popenta:
In this PR the class `BunchOfTransactions` has also been removed, as it was not used within `mxpy`.
As some people were referencing `mxpy` as a package just to use this class, the following code is an alternative implementation to that class, that can be copied into your own project and used directly.

```
from typing import Any, Dict, List, Optional, Protocol, Tuple

from multiversx_sdk_core import Transaction, TransactionPayload
from multiversx_sdk_wallet import UserSigner


class IAddress(Protocol):
    def bech32(self) -> str:
        ...


class ITransaction(Protocol):
    def to_dictionary(self) -> Dict[str, Any]:
        ...


class INetworkProvider(Protocol):
    def send_transactions(self, transactions: List[ITransaction]) -> Tuple[int, str]:
        ...


class BunchOfTransactions:
    def __init__(self) -> None:
        self.transactions: List[Transaction] = []

    def add_prepared(self, transaction: Transaction) -> None:
        self.transactions.append(transaction)

    def add(self,
            sender: UserSigner,
            receiver_address: IAddress,
            chain: str,
            gas_limit: int,
            nonce: int,
            value: int,
            data: str,
            gas_price: Optional[int] = None,
            version: Optional[int] = None,
            options: Optional[int] = None) -> None:
        tx = Transaction(
            chain_id=chain,
            sender=sender.get_pubkey().to_address("erd"),
            receiver=receiver_address,
            gas_limit=gas_limit,
            gas_price=gas_price,
            nonce=nonce,
            value=value,
            data=TransactionPayload.from_str(data),
            version=version,
            options=options
        )

        tx.signature = sender.sign(tx)
        self.transactions.append(tx)

    def send(self, proxy: INetworkProvider) -> Tuple[int, str]:
        logger.info(f"BunchOfTransactions.send: {len(self.transactions)} transactions")
        num_sent, hashes = proxy.send_transactions(self.transactions)
        return num_sent, hashes
```


multiversx_sdk_cli/transactions.py
```diff
@@ -323,3 +119,90 @@ def get_guardian_account_from_args(args: Any):
...
+    raise errors.KnownError("Took too long to get transaction.")
+
+
+def tx_to_dictionary_as_inner(tx: Transaction) -> Dict[str, Any]:
+    dictionary = tx.to_dictionary()
+    dictionary["receiver"] = base64.b64encode(bytes.fromhex(tx.receiver.hex())).decode()  # type: ignore
+    dictionary["sender"] = base64.b64encode(bytes.fromhex(tx.sender.hex())).decode()  # type: ignore
+    dictionary["chainID"] = base64.b64encode(tx.chainID.encode()).decode()
+    dictionary["signature"] = base64.b64encode(bytes(bytearray(tx.signature))).decode()
+    dictionary["value"] = tx.value
```
bogdan-rosianu, line 156: what about gas price / gas limit ?

multiversx_sdk_cli/transactions.py
```diff
@@ -323,3 +119,90 @@ def get_guardian_account_from_args(args: Any):
...
+def _dict_to_json(dictionary: Dict[str, Any]) -> bytes:
+    serialized = json.dumps(dictionary, separators=(',', ':')).encode("utf8")
+    return serialized
+
+
+def serialize_as_inner(tx: Transaction) -> str:
+    inner_dictionary = tx_to_dictionary_as_inner(tx)
+    serialized = _dict_to_json(inner_dictionary)
+    serialized_hex = serialized.hex()
+    return f"relayedTx@{serialized_hex}"
```
bogdan-rosianu, line 170: I think this function should only return the serialized tx, not the entire data field. otherwise, maybe rename the function?

multiversx_sdk_cli/transactions.py
popenta: `dictionary = tx.to_dictionary()` on the first line of the function. These are the values that need to be overwritten for the relayed transaction.

bogdan-rosianu (approved):

multiversx_sdk_cli/contracts.py
```diff
@@ -62,7 +62,7 @@ class IContractQueryResponse(Protocol):
 
 
 class SmartContract:
-    def __init__(self, address: Optional[Address] = None, bytecode=None, metadata=None):
+    def __init__(self, address: Optional[IAddress] = EmptyAddress(), bytecode=None, metadata=None):
```
andreibancioiu, line 65: I think `SmartContract` is the next in line, to be removed (replaced).

multiversx_sdk_cli/tests/test_accounts.py
```diff
@@ -1,44 +1,45 @@
...
-    transaction.receiver = "erd188nydpkagtpwvfklkl2tn0w6g40zdxkwfgwpjqc2a2m2n7ne9g8q2t22sr"
-    transaction.gasPrice = 200000000000000
-    transaction.gasLimit = 500000000
-    transaction.data = "foo"
-    transaction.chainID = "chainID"
-    transaction.version = 1
-    transaction.signature = alice.sign_transaction(transaction)
-
-    assert "0e69f27e24aba2f3b7a8842dc7e7c085a0bfb5b29112b258318eed73de9c8809889756f8afaa74c7b3c7ce20a028b68ba90466a249aaf999a1a78dcf7f4eb40c" == transaction.signature
+    transaction = Transaction(
```
andreibancioiu, line 14: I think these tests can be removed (the one with signing a transaction), since we have them (most of the flow, at least) in sdk-core, as well. Can stay for now, but we can remove them in the future.

multiversx_sdk_cli/transactions.py
```diff
@@ -323,3 +119,90 @@ def get_guardian_account_from_args(args: Any):
...
+
+
+def load_transaction_from_file(f: TextIO) -> Transaction:
+    data_json: bytes = f.read().encode()
+    fields = json.loads(data_json).get("tx") or json.loads(data_json).get("emittedTransaction")
+
+    instance = JSONTransaction()
+    instance.__dict__.update(fields)
```
andreibancioiu, line 178: Perhaps we could have used `Transaction.from_dictionary`?

https://github.com/multiversx/mx-sdk-py-core/blob/main/multiversx_sdk_core/transaction.py#L105

Then, we can remove `JSONTransaction` as well, I think. 

(for a future PR).

andreibancioiu (approved):

multiversx_sdk_cli/contracts.py
popenta: I agree. I don't know if we should do it now or wait for the sdk to be compliant with the specs.

multiversx_sdk_cli/tests/test_accounts.py
popenta: Yes, indeed. Will be removed in the future.

multiversx_sdk_cli/transactions.py
popenta: Indeed. I forgot we have that in sdk-core. Will be replaced in a future PR.

Issue #293, opened 2023-07-21
"mxpy contract report" - automatically install dependencies, if missing


andreibancioiu:
Suggested by @MWFIAE: https://t.me/MultiversXDevelopers/116434

popenta:
The issue has been solved in this [PR](https://github.com/multiversx/mx-sdk-py-cli/pull/302)

PR #302, opened 2023-08-08
Install twiggy if not installed for contract report

popenta:
When building a contract for `mxpy contract report` use `sc-meta` to build the contract, as well. If `twiggy` is not installed, it will be automatically installed.
Solves issue https://github.com/multiversx/mx-sdk-py-cli/issues/293.

### DIFF
--- a/multiversx_sdk_cli/projects/project_rust.py
+++ b/multiversx_sdk_cli/projects/project_rust.py
@@ -128,7 +128,7 @@ class ProjectRust(Project):
         return dependencies.get_module_by_key("rust").get_env()
 
     def build_wasm_with_debug_symbols(self, build_options: Dict[str, Any]):
-        cwd = self.get_meta_folder().parent
+        cwd = self.path
         env = self.get_env()
         target_dir: str = build_options.get("target-dir", "")
         target_dir = self._ensure_cargo_target_dir(target_dir)
--- a/multiversx_sdk_cli/projects/project_rust.py
+++ b/multiversx_sdk_cli/projects/project_rust.py
@@ -128,14 +128,14 @@ class ProjectRust(Project):
         return dependencies.get_module_by_key("rust").get_env()
 
     def build_wasm_with_debug_symbols(self, build_options: Dict[str, Any]):
-        cwd = self.get_meta_folder()
+        cwd = self.get_meta_folder().parent
         env = self.get_env()
-        target_dir: str = build_options.get("cargo-target-dir", "")
+        target_dir: str = build_options.get("target-dir", "")
         target_dir = self._ensure_cargo_target_dir(target_dir)
 
         args = [
-            "cargo",
-            "run",
+            "sc-meta",
+            "all",
             "build",
             "--wasm-symbols",
             "--wasm-suffix", "dbg",
--- a/multiversx_sdk_cli/projects/report/do_report.py
+++ b/multiversx_sdk_cli/projects/report/do_report.py
@@ -29,8 +29,7 @@ def _build_report(args: Any, build_options: Dict[str, Any]) -> None:
     project_paths = get_project_paths_recursively(base_path)
     options = get_default_report_features()
 
-    args_copy = copy.deepcopy(args)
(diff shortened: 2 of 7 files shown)


multiversx_sdk_cli/projects/report/report_creator.py
```diff
@@ -47,7 +52,7 @@ def _create_project_report(self, parent_path: Path, project_path: Path) -> Proje
         project = load_project(project_path)
 
         if not self.skip_build:
-            project.build(self.build_options)
+            project.build(self.build_args, self.build_options)
```
andreibancioiu, line 55: I think we can remove `get_option()` from `ProjectBase`, `ProjectRust` etc. If so, can we also remove the "options" property and possibly the build options args? (I did not completely check this).

multiversx_sdk_cli/projects/report/features/twiggy_paths_check.py
```diff
@@ -26,9 +27,22 @@ def requires_twiggy_paths(self):
...
 
+def ensure_twiggy_is_installed():
+    module = dependencies.get_module_by_key("twiggy")
+    default_tag: str = config.get_dependency_tag(module.key)
+
+    installed = module.is_installed(default_tag)
+
+    if not installed:
+        logger.warning("twiggy is not installed! Installing twiggy...")
+        dependencies.install_module("twiggy")
```
andreibancioiu, line 38: Let's check if this line alone would work, since it also performs install checks within its implementation. If so, then `ensure_twiggy_is_installed` can directly be: `dependencies.install_module("twiggy")`. Sorry if I misunderstood.

multiversx_sdk_cli/projects/report/do_report.py
```diff
@@ -22,14 +25,25 @@ def do_report(args: Any, build_options: Any) -> None:
 
 
 def _build_report(args: Any, build_options: Dict[str, Any]) -> None:
-    base_path = Path(args.project)
+    base_path = Path(args.path)
     project_paths = get_project_paths_recursively(base_path)
     options = get_default_report_features()
-    report_creator = ReportCreator(options, skip_build=args.skip_build, skip_twiggy=args.skip_twiggy, build_options=build_options)
+
+    args_copy = copy.deepcopy(args)
```
andreibancioiu, line 32: Can we move this within `_prepare_args_for_build` (thus, the preparation of args would be a bit more encapsulated - that is, the copying of the args is bit of a lower-level detail)? And `_prepare_args_for_build()` would actually return the new, copied & cleaned (filtered) object.

multiversx_sdk_cli/projects/project_rust.py
```diff
@@ -128,14 +128,14 @@ def get_env(self):
         return dependencies.get_module_by_key("rust").get_env()
 
     def build_wasm_with_debug_symbols(self, build_options: Dict[str, Any]):
-        cwd = self.get_meta_folder()
+        cwd = self.get_meta_folder().parent
```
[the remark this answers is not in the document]
popenta: Changed to `self.path`

multiversx_sdk_cli/projects/report/report_creator.py
popenta: I don't think we can. I deleted `get_options()` as it wasn't used but when building the contract we have the following line:
`with_wasm_opt = not self.options.get("no-wasm-opt")`. I think both `options` and `build_args` are needed.

multiversx_sdk_cli/projects/report/features/twiggy_paths_check.py
popenta: Works! Removed the `ensure_twiggy_is_installed()` function. Now using `dependencies.install_module("twiggy")`

multiversx_sdk_cli/projects/report/do_report.py
popenta: Done!

andreibancioiu (approved):

PR #319, opened 2023-08-28
Fix contract build in case of Rust error

popenta:
`mxpy contract build` would display `no build error` even if the Rust error was displayed in the terminal.

Fixes https://github.com/multiversx/mx-sdk-py-cli/issues/312

Also added unit tests.


multiversx_sdk_cli/utils.py
```diff
@@ -239,15 +239,6 @@ def query_latest_release_tag(repo: str) -> str:
     return latest_release_tag
```
andreibancioiu, line 241: Indeed, old, deprecated code.

andreibancioiu (approved):

camilbancioiu (approved):

PR #329, opened 2023-09-07
Fix no wasm opt argument

popenta:


bogdan-rosianu (dismissed):

danielailie (approved):

multiversx_sdk_cli/cli_contracts.py
```diff
@@ -463,7 +463,7 @@ def do_reproducible_build(args: Any):
     utils.ensure_folder(output_path)
 
     options = args.__dict__
-    no_wasm_opt = options.get("no-wasm-opt", True)
+    no_wasm_opt = options.get("no_wasm_opt", True)
```
andreibancioiu: For reproducible builds, if argument is not provided, it should default to `False` (wasm opt should be applied).

andreibancioiu (approved):

bogdan-rosianu (approved):

PR #339, opened 2023-09-25
Add url to transactions in the explorer

popenta:
Now, when sending transactions, the url to the explorer is computed and displayed in the terminal.

### DIFF
--- a/multiversx_sdk_cli/cli_shared.py
+++ b/multiversx_sdk_cli/cli_shared.py
@@ -21,6 +21,7 @@ from multiversx_sdk_cli.interfaces import ITransaction
 from multiversx_sdk_cli.ledger.ledger_functions import do_get_ledger_address
 from multiversx_sdk_cli.simulation import Simulator
 from multiversx_sdk_cli.transactions import send_and_wait_for_result
+from multiversx_sdk_cli.utils import log_explorer_transaction
 from multiversx_sdk_cli.ux import show_warning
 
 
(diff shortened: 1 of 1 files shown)


danielailie (approved):

multiversx_sdk_cli/cli_shared.py
```diff
@@ -274,8 +275,16 @@ def send_or_simulate(tx: ITransaction, args: Any, dump_output: bool = True) -> C
...
             output_builder.set_simulation_results(simulation)
     finally:
+        output_transaction = output_builder.build()
+
         if dump_output:
-            utils.dump_out_json(output_builder.build(), outfile=outfile)
+            utils.dump_out_json(output_transaction, outfile=outfile)
+
+        if send_only:
+            log_explorer_transaction(
```
andreibancioiu, line 284: In a future PR, we have to adjust the implementation of `log_explorer_transaction` to take `config` items (not existing yet) into consideration. E.g. some might want to use the `devnet2` URL for chain ID `D`.

https://github.com/multiversx/mx-sdk-py-cli/blob/main/multiversx_sdk_cli/utils.py#L246

andreibancioiu (approved):

multiversx_sdk_cli/cli_shared.py
andreibancioiu: Let's create a GitHub issue on this matter.

popenta: I've thought about this, but since in about one month there will be no `devnet2` I decided to leave it as it is.

PR #343, opened 2023-10-02
Message signing and verifying

popenta:
In this PR I've added two new sub-commands for the `wallet` command group:
- `mxpy wallet sign-message`
- `mxpy wallet verify-message`

### mxpy wallet sign-message

This command is used to sign a message. It requires the `--message` argument and you'll also need to pass in a wallet that will be used for signing the message.

When running `mxpy wallet sign-message --message test --pem alice.pem` the output will look something like this:
```
{
            "address": "erd1qyu5wthldzr8wx5c9ucg8kjagg0jfs53s8nr3zpz3hypefsdd8ssycr6th",
            "message": "test",
            "signature": "0x7aff43cd6e3d880a65033bf0a1b16274854fd7dfa9fe5faa7fa9a665ee851afd4c449310f5f1697d348e42d1819eaef69080e33e7652d7393521ed50d7427a0e"
}
```

### mxpy wallet verify-message

This command is used for verifying a previously signed message. It requires the `--address` argument which is the `bech32` address of the signer, the `--message` argument which represents the signed message(in readable format) and the `--signature` argument which is the message signature hex encoded.

To verify the message signed above we can run the following command:
```
mxpy wallet verify-message \
--address erd1qyu5wthldzr8wx5c9ucg8kjagg0jfs53s8nr3zpz3hypefsdd8ssycr6th \
--message test \
--signature 0x7aff43cd6e3d880a65033bf0a1b16274854fd7dfa9fe5faa7fa9a665ee851afd4c449310f5f1697d348e42d1819eaef69080e33e7652d7393521ed50d7427a0e
```

The output will look like this:
```
The message "test" was signed by erd1qyu5wthldzr8wx5c9ucg8kjagg0jfs53s8nr3zpz3hypefsdd8ssycr6th
```

In case the message was not signed by someone with that address the output will be:
```
The message "not signed" was NOT signed by erd1qyu5wthldzr8wx5c9ucg8kjagg0jfs53s8nr3zpz3hypefsdd8ssycr6th
```


multiversx_sdk_cli/tests/test_cli_wallet.py
```diff
@@ -239,6 +239,65 @@ def test_wallet_convert_pem_to_pubkey(capsys: Any):
     assert out == "0139472eff6886771a982f3083da5d421f24c29181e63888228dc81ca60d69e1"
 
 
+def test_wallet_sign_message(capsys: Any):
```
andreibancioiu, line 242: :rocket:

multiversx_sdk_cli/cli_wallet.py
```diff
@@ -78,6 +80,27 @@ def setup_parser(args: List[str], subparsers: Any) -> Any:
...
+    cli_shared.add_wallet_args(args, sub)
+    sub.set_defaults(func=sign_message)
+
+    sub = cli_shared.add_command_subparser(
+        subparsers,
+        "wallet",
+        "verify-message",
+        "Verify a previously message"
+    )
+    sub.add_argument("--bech32-address", required=True, help="the bech32 address of the signer")
```
andreibancioiu, line 99: Just `--address`, perhaps?

multiversx_sdk_cli/cli_wallet.py
```diff
@@ -247,3 +270,31 @@ def do_bech32(args: Any):
...
+def verify_signed_message(args: Any):
+    bech32_address = args.bech32_address
+    verifier = UserVerifier.from_address(Address.from_bech32(bech32_address))
+
+    message: str = args.message
+    verifiable_message = MessageV1.from_string(message)
+
+    signature: str = args.signature
+    if signature.startswith("0x"):
+        signature = signature[2:]
```
andreibancioiu, line 292: Indeed, best this way. Possibly handle `0X`, as well (uppercase X)?

multiversx_sdk_cli/cli_wallet.py
```diff
@@ -247,3 +270,31 @@ def do_bech32(args: Any):
...
+
+    verifiable_message.signature = bytes.fromhex(signature)
+    is_signed = verifier.verify(verifiable_message)
+
+    if is_signed:
+        show_message(f"""The message "{message}" was signed by {bech32_address}""")
```
andreibancioiu, line 298: Perhaps add a prefix for these messages, such as "success" and "failed" (as in "verification failed". E.g: 

```
SUCCESS: The messsage ...
```

Optional, we should ask for additional opinion on this.

multiversx_sdk_cli/message.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class SignableMessage:
+    def __init__(self, message: str, account: Account) -> None:
+        self.message = message
+        self.account = account
+
+    def sign(self) -> None:
+        hex_signature = self.account.sign_message(self.message.encode())
+        self.signature = bytes.fromhex(hex_signature)
```
andreibancioiu, line 13: The "signable" message now becomes a "signed message".

Alternatively, could be `SignedMessage(message, account)` and also do `self.signature = ...` in the constructor. Not an actual improvement, though. 

Or have this file renamed to `sign_verify.py` and have a `sign_message(message, account)` return a simple DTO called `SignedMessage` (also defined in this file), that has the `to_dictionary()` method, as well. Then, some parts of the verify logic can be moved here, too (for symmetry).

Optional, can also stay as it is.

multiversx_sdk_cli/cli_wallet.py
popenta: Done

multiversx_sdk_cli/cli_wallet.py
popenta: Done

multiversx_sdk_cli/cli_wallet.py
popenta: I have added prefixes, we can still ask for some other opinions, maybe wait to see what the second reviewer thinks about this.

multiversx_sdk_cli/message.py
popenta: Indeed, great suggestion!

andreibancioiu (approved):

ssd04 (approved):

andreibancioiu (approved):

PR #359, opened 2023-11-06
Integrate latest packages

popenta:
Integrated the latest versions of `sdk-core`, `sdk-wallet` and `sdk-network-providers` into mxpy.

### DIFF
--- a/multiversx_sdk_cli/contracts.py
+++ b/multiversx_sdk_cli/contracts.py
@@ -3,7 +3,7 @@ import logging
 from typing import Any, List, Optional, Protocol, Sequence, Tuple
 
 from multiversx_sdk_core import Transaction, TransactionPayload
-from multiversx_sdk_core.address import Address, compute_contract_address
+from multiversx_sdk_core.address import Address, AddressComputer
 from multiversx_sdk_network_providers.interface import IAddress, IContractQuery
 
 from multiversx_sdk_cli import config, constants, errors
@@ -69,7 +69,8 @@ class SmartContract:
 
     def deploy(self, owner: Account, arguments: List[Any], gas_price: int, gas_limit: int, value: int, chain: str, version: int, guardian: str, options: int) -> Transaction:
         self.owner = owner
-        self.address = compute_contract_address(self.owner.address, self.owner.nonce, DEFAULT_HRP)
+        address_computer = AddressComputer(number_of_shards=3)
+        self.address = address_computer.compute_contract_address(self.owner.address, self.owner.nonce)
 
         arguments = arguments or []
         gas_price = int(gas_price)
@@ -79,7 +80,7 @@ class SmartContract:
         tx = Transaction(
             chain_id=chain,
             sender=owner.address,
-            receiver=Address.from_bech32(ADDRESS_ZERO_BECH32),
+            receiver=Address.new_from_bech32(ADDRESS_ZERO_BECH32),
             gas_limit=gas_limit,
             gas_price=gas_price,
             nonce=owner.nonce,
--- a/multiversx_sdk_cli/contracts.py
+++ b/multiversx_sdk_cli/contracts.py
@@ -79,19 +79,19 @@ class SmartContract:
 
         tx = Transaction(
             chain_id=chain,
-            sender=owner.address,
-            receiver=Address.new_from_bech32(ADDRESS_ZERO_BECH32),
+            sender=owner.address.to_bech32(),
+            receiver=ADDRESS_ZERO_BECH32,
             gas_limit=gas_limit,
             gas_price=gas_price,
             nonce=owner.nonce,
-            value=value,
-            data=self.prepare_deploy_transaction_data(arguments),
+            amount=value,
+            data=self.prepare_deploy_transaction_data(arguments).data,
             version=version,
             options=options
         )
 
         if guardian:
-            tx.guardian = Address.from_bech32(guardian)
+            tx.guardian = guardian
 
         tx.signature = bytes.fromhex(owner.sign_transaction(tx))
         return tx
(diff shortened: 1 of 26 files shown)


multiversx_sdk_cli/contracts.py
```diff
@@ -78,19 +79,19 @@ def deploy(self, owner: Account, arguments: List[Any], gas_price: int, gas_limit
 
         tx = Transaction(
             chain_id=chain,
-            sender=owner.address,
-            receiver=Address.from_bech32(ADDRESS_ZERO_BECH32),
+            sender=owner.address.to_bech32(),
```
andreibancioiu, line 82: Here can we use the new factory instead? Can be done now or in a next PR.

andreibancioiu: Same for other features in `contracts.py`.

requirements.txt
```diff
@@ -9,6 +9,6 @@ semver
 requests-cache
 rich==13.3.4
 
-multiversx-sdk-core>=0.6.0,<0.7.0
-multiversx-sdk-network-providers>=0.11.0,<0.12.0
-multiversx-sdk-wallet>=0.8.0,<0.9.0
+multiversx-sdk-core==0.7.1
```
andreibancioiu, line 12: It's good to use fixed versions, since `mxpy` is an end-application (not a library) :+1: 

Plus the fact that it's always installed within a venv (at least, this is what mxpy-up does, this is what we recommend).

andreibancioiu: On the other hand, it's also nice to automatically benefit from hotfixes in dependencies (thus `>=` with `<` would have also worked).

multiversx_sdk_cli/transactions.py
```diff
@@ -148,12 +148,34 @@ def _send_transaction_and_wait_for_result(proxy: INetworkProvider, payload: ITra
 
 
 def tx_to_dictionary_as_inner(tx: Transaction) -> Dict[str, Any]:
-    dictionary = tx.to_dictionary()
-    dictionary["receiver"] = base64.b64encode(bytes.fromhex(tx.receiver.hex())).decode()  # type: ignore
-    dictionary["sender"] = base64.b64encode(bytes.fromhex(tx.sender.hex())).decode()  # type: ignore
-    dictionary["chainID"] = base64.b64encode(tx.chainID.encode()).decode()
-    dictionary["signature"] = base64.b64encode(bytes(bytearray(tx.signature))).decode()
-    dictionary["value"] = tx.value
+    dictionary: Dict[str, Any] = {}
```
andreibancioiu, line 151: Perhaps we can also rename the function to `tx_to_dictionary_as_inner_for_relayed_v1`. Or perhaps can we use the builder from core instead? Can be in a next PR.

pyproject.toml
```diff
@@ -27,9 +27,9 @@ dependencies = [
   "semver",
   "requests-cache",
   "rich==13.3.4",
-  "multiversx-sdk-network-providers>=0.11.0,<0.12.0",
-  "multiversx-sdk-wallet>=0.8.0,<0.9.0",
-  "multiversx-sdk-core>=0.6.0,<0.7.0"
+  "multiversx-sdk-network-providers==0.12.0",
```
andreibancioiu, line 30: `0.12.1`.

multiversx_sdk_cli/tests/test_cli_transactions.py
```diff
@@ -0,0 +1,32 @@
+from pathlib import Path
+from typing import Any
+
+from multiversx_sdk_cli.cli import main
+
+testdata_path = Path(__file__).parent / "testdata"
+
+
+def test_relayed_v1_transaction(capsys: Any):
```
andreibancioiu, line 9: :rocket:

multiversx_sdk_cli/tests/test_cli_transactions.py
```diff
@@ -0,0 +1,32 @@
...
+        "tx", "new",
+        "--pem", str(multi_user_pem),
+        "--pem-index", str(address_index),
+        "--receiver", "erd1qqqqqqqqqqqqqqqpqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqzllls8a5w6u",
+        "--nonce", "198",
+        "--gas-limit", "60000000",
+        "--data", "getContractConfig",
+        "--version", "1",
+        "--chain", "T",
+        "--relay"
```
andreibancioiu, line 23: In the future, we should add support for relayed V2, as well.

andreibancioiu (commented):
Very nice :tada:

multiversx_sdk_cli/contracts.py
popenta: Indeed. Was thinking about it, but I wanted to make sure everything works as expected. Will refactor to use the factory in a next PR.

requirements.txt
popenta: I'm a big fan of fixed versions, but I agree it's nice to benefit from hotfixes. Will indeed change to use `<`.

multiversx_sdk_cli/transactions.py
popenta: Renamed to `tx_to_dictionary_as_inner_for_relayed_v1`. Was thinking about using the builder, but since here we simply format the data field of the relayed transaction it's not suitable. We also marked the builders as deprecated, will probably use the factory, when it's ready.

multiversx_sdk_cli/tests/test_cli_transactions.py
popenta: Indeed

andreibancioiu (approved):

ssd04 (approved):

PR #366, opened 2023-11-27
Add extra checks when installing Rust - display some warning messages, if needed

andreibancioiu:
Display WARN messages if `CARGO_HOME` and `RUSTUP_HOME` are set before the installation:

https://rust-lang.github.io/rustup/installation/index.html#choosing-where-to-install 

Unset the variables if installing rust with `--overwrite`.

### DIFF
--- a/multiversx_sdk_cli/dependencies/modules.py
+++ b/multiversx_sdk_cli/dependencies/modules.py
@@ -273,7 +273,7 @@ class Rust(DependencyModule):
         return all(dependency is not None for dependency in dependencies)
 
     def install(self, overwrite: bool) -> None:
-        self._check_install_env()
+        self._check_install_env(apply_correction=overwrite)
 
         module = dependencies.get_module_by_key("rust")
         tag: str = config.get_dependency_tag(module.key)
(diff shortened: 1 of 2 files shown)


multiversx_sdk_cli/dependencies/modules.py
```diff
@@ -273,10 +273,13 @@ def is_installed(self, tag: str) -> bool:
         return all(dependency is not None for dependency in dependencies)
 
     def install(self, overwrite: bool) -> None:
+        self._check_install_env()
```
popenta, line 276: I think that if someone uses the `--overwrite` flag, the existing `CARGO_HOME` and `RUSTUP_HOME` should reset to the default location, as if `rust` was not previously installed.

andreibancioiu: Indeed, we should do that :raised_hands:

andreibancioiu: Fixed.

ovidiuolteanu (approved):

PR #368, opened 2023-12-05
Mxpy auto-update command

popenta:


popenta:
No longer needed since we can install and upgrade using **pipx**.
Done in [this PR](https://github.com/multiversx/mx-sdk-py-cli/pull/374).

PR #374, opened 2023-12-12
Install mxpy using pipx

popenta:
**mxpy** can now be installed using [**pipx**](https://pipx.pypa.io/stable/).

### DIFF
--- a/multiversx_sdk_cli/cli.py
+++ b/multiversx_sdk_cli/cli.py
@@ -2,6 +2,7 @@ import argparse
 import logging
 import sys
 from argparse import ArgumentParser
+from pathlib import Path
 from typing import Any, List
 
 from rich.logging import RichHandler
@@ -38,6 +39,7 @@ def main(cli_args: List[str] = sys.argv[1:]):
 
 
 def _do_main(cli_args: List[str]):
+    ensure_multiversx_sdk_directory_exists()
     argv_with_config_args = config.add_config_args(cli_args)
     parser = setup_parser(argv_with_config_args)
     args = parser.parse_args(argv_with_config_args)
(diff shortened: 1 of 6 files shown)


multiversx_sdk_cli/cli.py
```diff
@@ -53,6 +55,11 @@ def _do_main(cli_args: List[str]):
         args.func(args)
 
 
+def ensure_multiversx_sdk_directory_exists() -> None:
+    sdk_path = Path("~/multiversx-sdk").expanduser().resolve()
```
andreibancioiu, line 59: I think we have this as a constant in `workstation.py` or something similar.

multiversx_sdk_cli/config.py
```diff
@@ -249,6 +249,8 @@ def determine_final_args(argv: List[str], config_args: Dict[str, Any]) -> List[s
 def get_dependency_directory(key: str, tag: str) -> Path:
     parent_directory = get_dependency_parent_directory(key)
     if tag == 'latest':
+        if not parent_directory.is_dir():
+            return parent_directory / tag
```
andreibancioiu, line 253: Is this a change related to pipx etc.?

andreibancioiu (commented):
Let's also add some GitHub workflows such as `install-ubuntu-pipx.yml, `install-macos-pipx.yml`, similar to the existing ones that use `mxpy-uo`.

multiversx_sdk_cli/cli.py
popenta: Removed `ensure_multiversx_sdk_directory_exists()`. Now using `utils.ensure_folder(config.SDK_PATH)`.
Thanks for the suggestion.

multiversx_sdk_cli/config.py
popenta: Not really. Randomly tested the `mxpy deps check testwallets` command and there was an error when the dependency was not installed. Since the tag is `latest` it used to do `Path("~/multiversx-sdk/testwallets").iterdir()` and since the folder did not exist it threw an error on `os.listdir()` in `pathlib`.

popenta:
> Let's also add some GitHub workflows such as `install-ubuntu-pipx.yml`, `install-macos-pipx.yml`, similar to the existing ones that use `mxpy-up`.

Added two GH actions to install `mxpy` using `pipx` for both Ubuntu and MacOS.

andreibancioiu (approved):

bogdan-rosianu (approved):

Issue #371, opened 2023-12-10
Do not automatically install Rust on "contract build"


andreibancioiu:
Should overcome issues such as:
 - https://t.me/MultiversXDevelopers/121311
 - https://t.me/MultiversXDevelopers/121314

Should also ameliorate https://github.com/multiversx/mx-sdk-py-cli/issues/338, since the underlying error of not being able to install `sc-meta` will not be hidden anymore by the error that happens at the time of running `contract build`: `No such file or directory: sc-meta`.

Additionally, do not perform Rust installation / installation check on `mxpy contract templates`. Currently, the output of `mxpy contract templates is`:

```
Found extra arguments in mxpy.json. Final arguments: ['contract', 'templates']
INFO     projects.templates: Checking if the necessarry dependencies are installed.                           templates.py:35
╭───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ We recommend using rust nightly-2023-05-26. If you'd like to overwrite your current version please run `mxpy deps install │
│ rust --overwrite`.                                                                                                        │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
INFO     modules: install: key=rust, tag=nightly-2023-05-26, overwrite=False                                   modules.py:283
INFO     modules: which ...
INFO     myprocess: run_process: ['sc-meta', 'templates'], in folder: None                                    myprocess.py:12
INFO     myprocess: Successful run. Output:                                                                   myprocess.py:16
...
```

Instead, perhaps, do `ensure_rust()` or something similar, that only checks for rust, without installing it.

popenta:
**mxpy** does not automatically install Rust on any contract operation.
Fixed in PR https://github.com/multiversx/mx-sdk-py-cli/pull/376.

PR #376, opened 2023-12-14
Throw error if Rust is not installed for contract operations: new, build, clean

popenta:
Previously, if one tried to build a contract using `mxpy` and it did not have Rust installed, `mxpy` would install it on the spot. Now, it does not install it anymore, it throws an error saying:
```sh
Rust is not installed on your machine. Please run `mxpy deps install rust --overwrite` and try again.
```

This was done  to make it easier for people to debug in case the installation failed. If `sc-meta`'s dependencies we're not installed it would fail the installation, but still proceed to try the contract build.

Fixes issue https://github.com/multiversx/mx-sdk-py-cli/issues/371.

### DIFF
--- a/multiversx_sdk_cli/dependency_checker.py
+++ b/multiversx_sdk_cli/dependency_checker.py
@@ -1,0 +1,11 @@
+from multiversx_sdk_cli import config, errors, ux
+from multiversx_sdk_cli.dependencies.modules import Rust
+
+
+def check_if_rust_is_installed():
+    RUST_MODULE_KEY = "rust"
+    rust_module = Rust(RUST_MODULE_KEY)
+    if not rust_module.is_installed(""):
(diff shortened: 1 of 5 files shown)


multiversx_sdk_cli/dependency_checker.py
```diff
@@ -0,0 +1,11 @@
+from multiversx_sdk_cli import config, errors, ux
+from multiversx_sdk_cli.dependencies.modules import Rust
+
+
+def check_if_rust_is_installed():
+    RUST_MODULE_KEY = "rust"
+    rust_module = Rust(RUST_MODULE_KEY)
+    if not rust_module.is_installed(""):
+        tag = config.get_dependency_tag(RUST_MODULE_KEY)
```
andreibancioiu, line 9: If not required, perhaps use `""` instead?

andreibancioiu (approved):

PR #388, opened 2024-01-05
Fix "deps check all" command

popenta:


multiversx_sdk_cli/dependencies/modules.py
```diff
@@ -338,7 +321,7 @@ def _install_sc_meta(self):
         tag = config.get_dependency_tag("sc-meta")
         args = ["cargo", "install", "multiversx-sc-meta", "--locked"]
 
-        if tag != "latest":
+        if tag != "":
```
andreibancioiu, line 324: Can be `if not tag` (also below).

multiversx_sdk_cli/cli_deps.py
```diff
@@ -34,16 +36,63 @@ def install(args: Any):
...
-    module = dependencies.get_module_by_key(name)
-    tag_to_check: str = config.get_dependency_tag(module.key)
+
+    if name == "all":
+        all_dependencies = dependencies.get_all_deps()
+        missing_dependencies: List[Tuple[str, str]] = []
+
+        for dependency in all_dependencies:
+            tag_to_check: str = config.get_dependency_tag(dependency.key)
+            is_installed = check_module_is_installed(dependency, tag_to_check)
```
andreibancioiu, line 46: This does not handle `rust` the complete way, as done below.

multiversx_sdk_cli/cli_deps.py
```diff
@@ -34,16 +36,63 @@ def install(args: Any):
...
 
     installed = module.is_installed(tag_to_check)
-    if installed:
-        logger.info(f"[{name} {tag_to_check}] is installed.")
-        return
+    return installed
+
 
-    raise errors.DependencyMissing(name, tag_to_check)
+def _get_actual_installed_rust_version() -> str:
```
andreibancioiu, line 95: :+1:

multiversx_sdk_cli/cli_deps.py
```diff
@@ -34,16 +36,63 @@ def install(args: Any):
...
+    args = ["rustup", "default"]
```
andreibancioiu, line 96: Will throw if rust is not installed beforehand, and the error is not caught (sorry if I'm mistaken).

Perhaps do a try / catch here and return empty string if missing?

multiversx_sdk_cli/cli_deps.py
```diff
@@ -34,16 +36,63 @@ def install(args: Any):
...
+    output = run_process(args, dump_to_stdout=False)
+    return output.rstrip("\n")
```
andreibancioiu, line 98: Perhaps use `strip()` directly, in case other whitespace might occur in the future?

multiversx_sdk_cli/cli_deps.py
```diff
@@ -34,16 +36,63 @@ def install(args: Any):
...
+            tag_to_check: str = config.get_dependency_tag(dependency.key)
+            is_installed = check_module_is_installed(dependency, tag_to_check)
+
+            if not is_installed:
+                missing_dependencies.append((dependency.key, tag_to_check))
+
+        if len(missing_dependencies):
+            raise errors.DependenciesMissing(missing_dependencies)
+        return
+    if name == "rust":
```
andreibancioiu, line 54: Nice for having the check done in this new, better way.

multiversx_sdk_cli/cli_deps.py
andreibancioiu: Can we have it directly in `Rust` dependency module - `modules.py`, by overriding `is_installed()`?

multiversx_sdk_cli/cli_deps.py
```diff
@@ -34,16 +36,63 @@ def install(args: Any):
 
 def check(args: Any):
     name: str = args.name
-    module = dependencies.get_module_by_key(name)
-    tag_to_check: str = config.get_dependency_tag(module.key)
+
+    if name == "all":
```
andreibancioiu, line 40: After re-thinking about this, maybe we can drop `check all` and `install all`? On the long run, maybe it's better (less issues, less corner cases) if we do.

Let's double check with @ovidiuolteanu and @ccorcoveanu. I feel `deps install all` is quite exotic.

multiversx_sdk_cli/dependencies/modules.py
popenta: Replaced with `if tag:`

multiversx_sdk_cli/cli_deps.py
popenta: Fixed

multiversx_sdk_cli/cli_deps.py
popenta: Moved it in the `Rust` module.

multiversx_sdk_cli/cli_deps.py
popenta: It didn't trow any error as it was called only if `is_installed` was `True`.

multiversx_sdk_cli/cli_deps.py
popenta: Now using `strip()`.

multiversx_sdk_cli/cli_deps.py
popenta: This is now refactored.

multiversx_sdk_cli/dependencies/modules.py
```diff
@@ -253,7 +253,24 @@ def is_installed(self, tag: str) -> bool:
...
         dependencies = [which_rustc, which_cargo, which_sc_meta, which_wasm_opt, which_twiggy]
-        return all(dependency is not None for dependency in dependencies)
+        installed = all(dependency is not None for dependency in dependencies)
+
+        if installed:
+            actual_version_installed = self._get_actual_installed_version()
+
+            if tag in actual_version_installed:
+                logger.info(f"[{self.key} {tag}] is installed.")
+            elif "Command 'rustup' not found" in actual_version_installed:
```
andreibancioiu, line 263: The error message might look differently on MacOS. E.g.

```
multiversx@papple ~ % rustupfoobar
zsh: command not found: rustupfoobar
multiversx@papple ~ % 
```

Thus, we can check for `"not found" in actual_version_installed` perhaps?

multiversx_sdk_cli/cli_deps.py
```diff
@@ -34,16 +35,38 @@ def install(args: Any):
...
+
+        if len(missing_dependencies):
+            raise errors.DependenciesMissing(missing_dependencies)
+        return
+    else:
+        module = dependencies.get_module_by_key(name)
+        tag_to_check: str = config.get_dependency_tag(module.key)
+
+        is_installed = check_module_is_installed(module, tag_to_check)
+        if is_installed and name != "rust":
```
andreibancioiu, line 58: For Rust, we don't display this log info? Or without this `if` the message will be duplicated by the one from `modules.py`? I think better to have it duplicated on the CLI than to have this custom `if` here (e.g. in the future we will forget its reason).

multiversx_sdk_cli/dependencies/modules.py
popenta: Indeed. Now the `elif` statement is as above.

multiversx_sdk_cli/cli_deps.py
popenta: Without the `if` statement the message will be duplicated by the one in `modules.py`. Why I added it it's because if you have a different Rust version other than the recommended one, you'll see a warning that the versions don't match but the log will show that the correct version is installed since it takes the tag from the config.
A problem that could occur with the other dependencies as well, now that I think about it, but is less common.

andreibancioiu (dismissed):

multiversx_sdk_cli/cli_deps.py
```diff
@@ -34,16 +35,38 @@ def install(args: Any):
 
 def check(args: Any):
     name: str = args.name
-    module = dependencies.get_module_by_key(name)
-    tag_to_check: str = config.get_dependency_tag(module.key)
+
+    if name == "all":
```
ssd04, line 39: `cargo` does not seem to be loaded correctly when trying to install all deps

ssd04: tested and it seems to work now :+1:

ssd04 (approved):

andreibancioiu (approved):

PR #406, opened 2024-01-31
Add shell completions

popenta:
`mxpy` is now capable of shell completions using [argcomplete](https://kislyuk.github.io/argcomplete/). Keep in mind that `argcomplete` works only for `bash` and `zsh` shells.

After installing `mxpy` using `pipx` the following line needs to be added to your shell file. For `bash`, add it in the `.bashrc` file.
```sh
eval "$(register-python-argcomplete mxpy)"
```
Then, open a new terminal and enjoy autocompletions. Press `<Tab>` for completions and `<Tab><Tab>` to see all the options. Don't forget that for a more detailed look the `--help` or `-h` arguments can be used.

For reference: https://stackoverflow.com/questions/14597466/custom-tab-completion-in-python-argparse/15289025#15289025
Another alternative was to use [shtab](https://github.com/iterative/shtab).

### DIFF
--- a/multiversx_sdk_cli/config.py
+++ b/multiversx_sdk_cli/config.py
@@ -202,22 +202,9 @@ def add_config_args(argv: List[str]) -> List[str]:
     except KeyError:
         return argv
 
-    check_for_deprecated_args(config_args)
-
     final_args = determine_final_args(argv, config_args)
     print(f"Found extra arguments in mxpy.json. Final arguments: {final_args}")
     return final_args
-
-
-def check_for_deprecated_args(args: List[str]) -> None:
-    if "proxy" in args:
-        show_warning("Providing `proxy` in the configuration file is deprecated. It will not be used. Please remove it!")
-
-    if "chainID" in args:
-        show_warning("Providing `chainID` in the configuration file is deprecated. It will not be used. Please remove it!")
-
-    if "txVersion" in args:
-        show_warning("Providing `txVersion` in the configuration file is deprecated. It will not be used. Please remove it!")
 
 
 def determine_final_args(argv: List[str], config_args: Dict[str, Any]) -> List[str]:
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -8,6 +8,7 @@ ledgercomm[hid]
 semver
 requests-cache
 rich==13.3.4
+argcomplete==3.2.2
```
andreibancioiu, line 11: Version published in 23rd of January 2024 :+1:

multiversx_sdk_cli/config.py
```diff
@@ -202,24 +202,11 @@ def add_config_args(argv: List[str]) -> List[str]:
...
         return argv
 
-    check_for_deprecated_args(config_args)
-
     final_args = determine_final_args(argv, config_args)
     print(f"Found extra arguments in mxpy.json. Final arguments: {final_args}")
     return final_args
 
 
-def check_for_deprecated_args(args: List[str]) -> None:
```
andreibancioiu, line 212: Actually, before an actual removal, let's convert them to errors - `throw new KnownError ...`. People should not be using them anymore.

andreibancioiu (commented):
Do people need to install `argcomplete` separately, so that  `register-python-argcomplete` is available? Since pipx will only install it in a venv.

popenta:
> Do people need to install `argcomplete` separately, so that `register-python-argcomplete` is available? Since pipx will only install it in a venv.
> 
> Is it possible to test this somehow in a github workflow (maybe not)?

No, they don't need to install argcomplete separately. I'll think of a way to test it.

multiversx_sdk_cli/config.py
popenta: I've deleted this because we have another function called `verify_deprecated_entries_in_config_file()` that was performing the same check.

andreibancioiu: Great!

andreibancioiu (approved):

acristurean (approved):

PR #408, opened 2024-02-05
Removed vmtools and contract test command

popenta:
`vmtools` is not needed anymore because `sc-meta` can now install all the required dependencies. It is now recommended to use `sc-meta` for creating, building and testing contracts and use `mxpy` to deploy, call and upgrade the contract.


andreibancioiu (approved):

ovidiuolteanu (approved):

popenta:
As discussed, we won't do this right now, we will keep the `contract test` command for now. Will close this PR as it's not needed.

PR #411, opened 2024-02-12
Modify contract test command to use sc-meta

popenta:
**mxpy** does not take care of `vmtools` anymore. The `vmtools` dependency has been removed.
Now, **mxpy** simply forwards the arguments to `sc-meta`. When calling `mxpy contract test`, the `sc-meta test` command is invoked. One could also use directly `sc-meta test` since `sc-meta` is installed when calling `mxpy deps install rust`.

The parameters have been changed to match the parameters needed by `sc-meta`.

`mxpy contract test` now uses the following parameters:
```
--path PATH  the directory of the contract (default: cwd)
--go         this arg runs rust and go tests (default: false)
--scen       this arg runs scenarios (default: false). If `--scen` and `--go` are both specified, scen overrides the go argument
--nocapture  this arg prints the entire output of the vm (default: false)
```


multiversx_sdk_cli/dependencies/modules.py
```diff
@@ -359,6 +310,17 @@ def _install_twiggy(self):
 
         myprocess.run_process(args, env=self.get_cargo_env())
 
+    def _install_sc_meta_deps(self):
+        # this is needed for sc-meta to run the tests
+        # if os is Windows skip insallation
```
andreibancioiu, line 315: Typo.

multiversx_sdk_cli/dependencies/modules.py
```diff
@@ -359,6 +310,17 @@ def _install_twiggy(self):
...
+        os = platform.system().lower()
+        if os == "windows":
+            return
```
andreibancioiu, line 318: Perhaps we should display a WARN?

danielailie (approved):

multiversx_sdk_cli/dependencies/modules.py
popenta: fixed

multiversx_sdk_cli/dependencies/modules.py
popenta: added a warning

andreibancioiu (approved):

danielailie (approved):

PR #418, opened 2024-03-12
Implemented a native auth client and faucet functionality

popenta:
Use `mxpy faucet request` to test the faucet functionality. It also requires a wallet and the chain id of the network you wish to receive funds on.

```bash
mxpy faucet request --pem alice.pem --chain D
```


multiversx_sdk_cli/cli_faucet.py
```diff
@@ -0,0 +1,76 @@
...
+    TESTNET = "https://testnet-api.multiversx.com"
+
+
+def setup_parser(args: List[str], subparsers: Any) -> Any:
+    parser = cli_shared.add_group_subparser(subparsers, "faucet", "Get xEGLD on Devnet or Testnet")
+    subparsers = parser.add_subparsers()
+
+    sub = cli_shared.add_command_subparser(subparsers, "faucet", "request", "Request xEGLD.")
+    cli_shared.add_wallet_args(args, sub)
+    sub.add_argument("--chain", required=True, help="the chain identifier")
```
ssd04, line 32: here it lists `--pem, --keyfile, --ledger, --chain` as required, but only one of the first three are required, right?

multiversx_sdk_cli/cli_faucet.py
```diff
@@ -0,0 +1,76 @@
...
+        signature=signature
+    )
+
+    logger.info(f"Requesting funds for address: {account.address.to_bech32()}")
+    call_web_Wallet_faucet(wallet_url=wallet, access_token=access_token)
+
+
+def call_web_Wallet_faucet(wallet_url: str, access_token: str):
+    faucet_url = f"{wallet_url}/faucet?accessToken={access_token}"
+    webbrowser.open_new_tab(faucet_url)
```
ssd04, line 64: i'm thinking of the main benefits of having this command from cli, since we still have to check in the browser

Dockerfile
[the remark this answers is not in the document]
popenta: it was committed by accident. removed!

multiversx_sdk_cli/native_auth_client.py
```diff
@@ -0,0 +1,101 @@
...
+
+from multiversx_sdk_cli.errors import NativeAuthClientError
+
+
+class NativeAuthClientConfig:
+    def __init__(
+            self,
+            origin: str = '',
+            api_url: str = "https://api.multiversx.com",
+            expiry_seconds: int = 60 * 60 * 24,
```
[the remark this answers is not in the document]
popenta: made some constants for that values

multiversx_sdk_cli/cli_faucet.py
popenta: we've had people asking about it and i think it's a little bit easier to use the cli instead of doing it manually even though you have to pass the recaptcha and then press the request button. This is just an initial implementation/poc. We were thinking about doing it another way, so the user does not have to do anything else but type the command but we have not yet found a non-exploitable way to do it and we'll also need some help from the colleagues working on the api.

multiversx_sdk_cli/native_auth_client.py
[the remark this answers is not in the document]
ssd04: here I was referring for these 2 variables: default api url and expiry time in seconds; i think it can be only one var for expiry_seconds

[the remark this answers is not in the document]
popenta: indeed, makes sense. done!

multiversx_sdk_cli/tests/test_native_auth_client.py
```diff
@@ -0,0 +1,128 @@
+from typing import Any, List
+
+import pytest
+
+from multiversx_sdk_cli.native_auth_client import (NativeAuthClient,
+                                                   NativeAuthClientConfig)
+
+
+def mock(mocker: Any, code: int, response: Any):
```
andreibancioiu, line 9: Nice :rocket:

multiversx_sdk_cli/tests/test_native_auth_client.py
```diff
@@ -0,0 +1,128 @@
...
+        token = client.initialize()
+        assert token == self.TOKEN
+
+    def test_generate_access_token(self):
+        client = NativeAuthClient()
+        access_token = client.get_token(self.ADDRESS, self.TOKEN, self.SIGNATURE)
+        assert access_token == self.ACCESS_TOKEN
+
+
+class TestNativeAuthWithGateway:
```
andreibancioiu, line 69: Both cases, nice coverage of flows :raised_hands:

multiversx_sdk_cli/native_auth_client.py
```diff
@@ -0,0 +1,104 @@
+import base64
+import json
+from typing import Any, Dict, Optional
+
+import requests
+
+from multiversx_sdk_cli.errors import NativeAuthClientError
+
+EXPIRY_TIME_IN_SECONDS = 60 * 60 * 2
```
andreibancioiu, line 9: Can also have prefix `DEFAULT_`.

multiversx_sdk_cli/native_auth_client.py
```diff
@@ -0,0 +1,104 @@
...
+    def _get_current_block_hash_using_api_fallback(self) -> str:
+        url = f"{self.config.api_url}/blocks?size=1&fields=hash"
+
+        if self.config.block_hash_shard:
+            url += f"&shard={self.config.block_hash_shard}"
+
+        response = self._execute_request(url)
+        return response[0]["hash"]
+
+    def encode_value(self, string: str) -> str:
```
andreibancioiu, line 94: Does this help?

https://docs.python.org/3/library/base64.html#base64.urlsafe_b64encode

`=` would have to be trimmed, nonetheless.

multiversx_sdk_cli/native_auth_client.py
```diff
@@ -0,0 +1,104 @@
...
+        self.block_hash_shard = block_hash_shard
+        self.gateway_url = gateway_url
+        self.extra_request_headers = extra_request_headers
+
+
+class NativeAuthClient:
+    def __init__(self, config: Optional[NativeAuthClientConfig] = None) -> None:
+        self.config = config or NativeAuthClientConfig()
+
+    def get_token(self, address: str, token: str, signature: str) -> str:
```
andreibancioiu, line 35: Functions can be re-ordered to match the usual flow. E.g. first "initialize", then ...

multiversx_sdk_cli/native_auth_client.py
```diff
@@ -0,0 +1,104 @@
...
+        if self.config.gateway_url:
+            return self._get_current_block_hash_using_gateway()
+        return self._get_current_block_hash_using_api()
+
+    def _get_current_block_hash_using_gateway(self) -> str:
+        round = self._get_current_round()
+        url = f"{self.config.gateway_url}/blocks/by-round/{round}"
+        response = self._execute_request(url)
+        blocks = response["data"]["blocks"]
+        block = [b for b in blocks if b["shard"] == self.config.block_hash_shard][0]
```
andreibancioiu, line 58: A bit fragile - it's possible to have a round where not all shards propose a block (missed block).

andreibancioiu: However, in the context of mxpy, we don't provide the shard.

andreibancioiu (dismissed):

multiversx_sdk_cli/native_auth_client.py
popenta: added

multiversx_sdk_cli/native_auth_client.py
popenta: Might help, but indeed `=` would have to be trimmed as well. Will leave it as it is for now.

ssd04 (approved):

PR #420, opened 2024-04-17
Raise error when given string argument without prefix

popenta:
The `mxpy contract` sub-commands now raise an error if a `string` argument was provided without the `str:` prefix.

### DIFF
--- a/multiversx_sdk_cli/tests/test_cli_contracts.py
+++ b/multiversx_sdk_cli/tests/test_cli_contracts.py
@@ -301,7 +301,7 @@ def test_contract_commands_argument_parameter():
         "--nonce", "7",
         "--chain", "D",
         "--gas-limit", "5000000",
-        "--arguments", "invalidargument",
+        "--arguments", "foobar",
     ])
     assert return_code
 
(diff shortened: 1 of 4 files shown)


multiversx_sdk_cli/tests/test_cli_contracts.py
```diff
@@ -290,6 +290,33 @@ def test_contract_deploy_without_required_arguments():
...
+    assert return_code
+
+    return_code = main([
+        "contract", "deploy",
+        "--bytecode", adder,
+        "--pem", alice,
+        "--nonce", "7",
+        "--chain", "D",
+        "--gas-limit", "5000000",
+        "--arguments", "str:invalidargument",
```
andreibancioiu, line 315: Maybe `str:foobar` (or something that doesn't say `invalid argument`).

popenta: Indeed. Changed.

ssd04 (approved):

andreibancioiu (approved):

PR #435, opened 2024-07-10
Support for abi files for contract operations

popenta:
Now, the abi file of the contract can be passed and arguments can be loaded from a json file that respects the format.

The args should all be placed in a list, so it should look something like this:
```json
[
    5,
    {
         { "bech32": "erd1..." }
    }
]
```


multiversx_sdk_cli/contracts.py
```diff
@@ -90,7 +93,9 @@ def prepare_deploy_transaction(self,
                                    version: int,
                                    options: int,
                                    guardian: str) -> Transaction:
-        args = prepare_args_for_factory(arguments) if arguments else []
+        args = arguments if arguments else []
+        if not args_from_file:
```
andreibancioiu, line 97: Maybe rename to `should_prepare_args` or `are_args_already_prepared` or something similar?

multiversx_sdk_cli/cli_contracts.py
```diff
@@ -232,6 +241,8 @@ def _add_arguments_arg(sub: Any):
     sub.add_argument("--arguments", nargs='+',
                      help="arguments for the contract transaction, as [number, bech32-address, ascii string, "
                      "boolean] or hex-encoded. E.g. --arguments 42 0x64 1000 0xabba str:TOK-a1c2ef true erd1[..]")
+    sub.add_argument("--arguments-file", type=str, help="a json file containing the arguments. ONLY if abi file is provided. "
+                     "E.g. { 'to': 'erd1...', 'amount': 10000000000 }")
```
andreibancioiu, line 245: The example in `e.g.` is not completely precise (the wrapping array is missing). Maybe point towards mxpy's cookbook (official docs)?

multiversx_sdk_cli/cli_contracts.py
```diff
@@ -433,6 +457,18 @@ def query(args: Any):
     utils.dump_out_json(result)
 
 
+def _get_contract_arguments(args: Any) -> Tuple[List[Any], bool]:
+    json_args = json.loads(Path(args.arguments_file).expanduser().read_text()) if args.arguments_file else None
+
+    if json_args and args.arguments:
+        raise Exception("Both '--arguments' and '--arguments-file' provided.")
```
andreibancioiu, line 464: Can be rephrase as suggestion for correction e.g. "provide either ... or ...".

andreibancioiu (dismissed):
:rocket: :rocket: :rocket:

multiversx_sdk_cli/cli_contracts.py
```diff
@@ -433,6 +457,18 @@ def query(args: Any):
     utils.dump_out_json(result)
 
 
+def _get_contract_arguments(args: Any) -> Tuple[List[Any], bool]:
```
axenteoctavian, line 460: Don't you need to check if `--abi` is provided in this function?

multiversx_sdk_cli/cli_contracts.py
axenteoctavian, line 1: I'm not sure how difficult would be to do the following:
- use `--arguments` and put there a json string instead of a file.
- use `--arguments` and put there some values (same as before) and encode from those values, not a json

multiversx_sdk_cli/contracts.py
popenta: renamed to `should_prepare_args`.

multiversx_sdk_cli/cli_contracts.py
popenta: indeed, added the wrapping array. we don't have anything in the cookbook regarding this topic (as far as I know)

multiversx_sdk_cli/cli_contracts.py
popenta: rephrased

multiversx_sdk_cli/cli_contracts.py
popenta: indeed. added a check for the abi if `--arguments-file` is provided.

multiversx_sdk_cli/cli_contracts.py
popenta: I've thought about having the args as a json string but that seems more complicated and I think it's easier to write them in a file. 
For the second bullet, that is already possible, we also have a test for that. If the abi is provided it will be passed to the factory that takes care of creating the transaction. The `--arguments-file` should be used only for complex arguments.

axenteoctavian (approved):

andreibancioiu (approved):

PR #436, opened 2024-07-11
Added suport for contract queries using abi file

popenta:
The same arguments as for contract calls have been added. The args are `--abi` and `--arguments-file`. Keep in mind the args should be placed inside a `list` like such:
```json
[
    5,
    {
         { "bech32": "erd1..." }
    }
]
```

When performing queries without the abi file the response looks like this:
```sh
[
    "0e"
]
```

But when using the abi file the response is parsed, and it looks like this:
```sh
[
    14
]
```

There are small differences in how values are returned. For example, if an endpoint returns `variadic<Address>`, when querying without abi, the response looks like this:
```
[
    "3c35d7eacb1a99a4a0ccaa537ef0cc0ae109f41f6c2b402203e633bfbb071f5c",
    "0139472eff6886771a982f3083da5d421f24c29181e63888228dc81ca60d69e1"
]
```

But when the abi is used, the response looks like this:
```
[
    [
        "3c35d7eacb1a99a4a0ccaa537ef0cc0ae109f41f6c2b402203e633bfbb071f5c",
        "0139472eff6886771a982f3083da5d421f24c29181e63888228dc81ca60d69e1"
    ]
]
```

### DIFF
--- a/multiversx_sdk_cli/cli_contracts.py
+++ b/multiversx_sdk_cli/cli_contracts.py
@@ -444,7 +444,7 @@ def upgrade(args: Any):
 def query(args: Any):
     logger.debug("query")
 
-    # workaround so we can use the function bellow to set chainID
+    # workaround so we can use the function below to set chainID
     args.chain = ""
     cli_shared.prepare_chain_id_in_args(args)
 
--- a/multiversx_sdk_cli/cli_contracts.py
+++ b/multiversx_sdk_cli/cli_contracts.py
@@ -452,7 +452,7 @@ def query(args: Any):
     abi = Abi.load(Path(args.abi)) if args.abi else None
     contract = SmartContract(config, abi)
 
-    arguments, args_from_file = _get_contract_arguments(args)
+    arguments, should_prepare_args = _get_contract_arguments(args)
     contract_address = Address.new_from_bech32(args.contract)
 
     proxy = ProxyNetworkProvider(args.proxy)
@@ -463,7 +463,7 @@ def query(args: Any):
         proxy=proxy,
         function=function,
         arguments=arguments,
-        args_from_file=args_from_file
+        should_prepare_args=should_prepare_args
     )
 
     utils.dump_out_json(result)
--- a/multiversx_sdk_cli/cli_contracts.py
+++ b/multiversx_sdk_cli/cli_contracts.py
@@ -13,7 +13,7 @@ from multiversx_sdk_cli.cli_output import CLIOutputBuilder
 from multiversx_sdk_cli.constants import NUMBER_OF_SHARDS
 from multiversx_sdk_cli.contract_verification import \
     trigger_contract_verification
-from multiversx_sdk_cli.contracts import SmartContract, query_contract
+from multiversx_sdk_cli.contracts import SmartContract
 from multiversx_sdk_cli.cosign_transaction import cosign_transaction
 from multiversx_sdk_cli.dependency_checker import check_if_rust_is_installed
 from multiversx_sdk_cli.docker import is_docker_installed, run_docker
@@ -132,6 +132,7 @@ def setup_parser(args: List[str], subparsers: Any) -> Any:
     sub = cli_shared.add_command_subparser(subparsers, "contract", "query",
                                            "Query a Smart Contract (call a pure function)")
     _add_contract_arg(sub)
+    _add_contract_abi_arg(sub)
     cli_shared.add_proxy_arg(sub)
     _add_function_arg(sub)
     _add_arguments_arg(sub)
@@ -443,17 +444,28 @@ def upgrade(args: Any):
 def query(args: Any):
     logger.debug("query")
 
-    # workaround so we can use the function bellow
+    # workaround so we can use the function bellow to set chainID
     args.chain = ""
     cli_shared.prepare_chain_id_in_args(args)
 
+    config = TransactionsFactoryConfig(args.chain)
+    abi = Abi.load(Path(args.abi)) if args.abi else None
+    contract = SmartContract(config, abi)
+
+    arguments, args_from_file = _get_contract_arguments(args)
     contract_address = Address.new_from_bech32(args.contract)
 
     proxy = ProxyNetworkProvider(args.proxy)
     function = args.function
-    arguments: List[Any] = args.arguments or []
 
-    result = query_contract(contract_address, proxy, function, arguments)
+    result = contract.query_contract(
+        contract_address=contract_address,
+        proxy=proxy,
+        function=function,
+        arguments=arguments,
+        args_from_file=args_from_file
+    )
+
     utils.dump_out_json(result)
 
 
--- a/multiversx_sdk_cli/cli_wallet.py
+++ b/multiversx_sdk_cli/cli_wallet.py
@@ -3,7 +3,7 @@ import json
 import logging
 import sys
 from pathlib import Path
-from typing import Any, List, Optional, Tuple, cast
+from typing import Any, List, Optional, Tuple
 
 from multiversx_sdk import (Address, Mnemonic, UserPEM, UserSecretKey,
                             UserWallet)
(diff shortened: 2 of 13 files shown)


multiversx_sdk_cli/contracts.py
```diff
@@ -192,6 +198,40 @@ def prepare_upgrade_transaction(self,
...
+
+        query = sc_query_controller.create_query(
+            contract=contract_address.to_bech32(),
+            function=function,
+            arguments=args
+        )
+
+        response = sc_query_controller.run_query(query)
+
+        if self._abi:
```
axenteoctavian, line 222: This if-else is not consistent.
a)
With ABI you return the data-parts
Without ABI your return other stuff + data-parts

b) errors for bad queries
With ABI you `raise SmartContractQueryError`
Without ABI you return the Dict with error message

1/ What is the reason you return different types?
You can just return `sc_query_controller.parse_query_response(response)` but for NO abi case return parts as hex.
2/ Please add some unit tests for bad queries (wrong --function or whatever)
3/ I can't find test for contract query with abi

popenta: Now, the `if-else` is gone. The querying is done in a `try-except` block so all the errors thrown are of type 
 `QueryContractError`. Did a small "fix" for the json encoder to convert all bytes to hex. Also added tests.

multiversx_sdk_cli/utils.py
```diff
@@ -37,9 +37,13 @@ def to_json(self):
 
 
 class BasicEncoder(json.JSONEncoder):
-    def default(self, o: Any):
+    def default(self, o: Any) -> Any:
         if isinstance(o, ISerializable):
             return o.to_dictionary()
+        if isinstance(o, bytes):
```
andreibancioiu, line 43: Generally speaking, is not a breaking change (e.g. it does not alter the output for existing flows that use --outfile or so).

multiversx_sdk_cli/utils.py
```diff
@@ -37,9 +37,13 @@ def to_json(self):
...
+            return o.hex()
+        if isinstance(o, list):
```
andreibancioiu, line 45: Is this necessary? (I did not do an extra check).

multiversx_sdk_cli/contracts.py
```diff
@@ -202,7 +201,7 @@ def query_contract(self,
                        contract_address: IAddress,
                        proxy: INetworkProvider,
                        function: str,
-                       arguments: List[Any],
+                       arguments: Union[List[Any], None],
```
andreibancioiu, line 204: Can also be `Optional` - or maybe not yet (older Python)?

multiversx_sdk_cli/contracts.py
```diff
@@ -211,26 +210,16 @@ def query_contract(self,
...
-        )
-
-        response = sc_query_controller.run_query(query)
-
-        if self._abi:
-            return sc_query_controller.parse_query_response(response)
-        else:
-            return [self._query_response_to_dict(response)]
+        try:
+            response = sc_query_controller.query(
```
andreibancioiu, line 214: So we don't support the `caller` and `callValue` parameters? But we didn't support them before anyway. I think we should, in the future.

andreibancioiu (dismissed):

multiversx_sdk_cli/contracts.py
axenteoctavian: Looks good to me

multiversx_sdk_cli/cli_contracts.py
```diff
@@ -443,17 +444,28 @@ def upgrade(args: Any):
 def query(args: Any):
     logger.debug("query")
 
-    # workaround so we can use the function bellow
+    # workaround so we can use the function bellow to set chainID
```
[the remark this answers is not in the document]
popenta: fixed

multiversx_sdk_cli/cli_wallet.py
```diff
@@ -134,6 +134,9 @@ def wallet_new(args: Any):
     else:
         mnemonic = Mnemonic.generate()
 
+    # this is done to get rid of the Pylance error: possibly unbound
+    mnemonic = cast(Mnemonic, mnemonic)  # type: ignore
```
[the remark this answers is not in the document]
popenta: indeed, extracted to a separate function.

multiversx_sdk_cli/dns.py
```diff
@@ -19,22 +20,56 @@ class INetworkProvider(Protocol):
...
+        args_from_file=False
+    )[0]
+
+    return_code = response["returnCode"]
+    if return_code == "ok":
+        print(f"name [{name}] is valid")
+    else:
+        print(f"name [{name}] is invalid")
+
+    print(response)
```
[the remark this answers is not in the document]
popenta: indeed, it will be documented in the release notes

multiversx_sdk_cli/dns.py
```diff
@@ -19,22 +20,56 @@ class INetworkProvider(Protocol):
...
 
 def resolve(name: str, proxy: INetworkProvider) -> Address:
     name_arg = "0x{}".format(str.encode(name).hex())
     dns_address = dns_address_for_name(name)
 
-    result = query_contract(dns_address, proxy, "resolve", [name_arg])
-    if len(result) == 0:
+    chain_id = proxy.get_network_config().chain_id
+    config = TransactionsFactoryConfig(chain_id)
+    contract = SmartContract(config)
```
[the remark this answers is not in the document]
popenta: done

multiversx_sdk_cli/contracts.py
```diff
@@ -63,6 +65,10 @@ class IContractQueryResponse(Protocol):
     return_data: List[str]
     return_code: str
     return_message: str
+    gas_used: int
```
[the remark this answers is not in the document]
popenta: I don't think we can drop it.

multiversx_sdk_cli/utils.py
popenta: from what I've tested it's not. deleted it.

multiversx_sdk_cli/contracts.py
popenta: changed to `Optional[List[Any]]`.

multiversx_sdk_cli/contracts.py
popenta: we can do that in the future, we didn't support them before.

multiversx_sdk_cli/cli_wallet.py
```diff
@@ -115,27 +115,7 @@ def wallet_new(args: Any):
...
-                break
-
-        if not is_wallet_generated:
-            raise WalletGenerationError(f"Couldn't generate wallet in shard {shard}")
-    else:
-        mnemonic = Mnemonic.generate()
-
-    # this is done to get rid of the Pylance error: possibly unbound
-    mnemonic = cast(Mnemonic, mnemonic)  # type: ignore
+    mnemonic = _generate_mnemonic_with_shard_constraint(shard)
```
andreibancioiu, line 118: Could be: `if shard is not none, then call the new function, else call Mnemonic.generate()`.

andreibancioiu (dismissed):

multiversx_sdk_cli/cli_wallet.py
popenta: done

andreibancioiu (approved):

multiversx_sdk_cli/dns.py
```diff
@@ -105,17 +97,14 @@ def name_hash(name: str) -> bytes:
...
-
-    response = contract.query_contract(
+    response = _query_contract(
         contract_address=dns_address,
         proxy=proxy,
-        function="getRegistrationCost",
-        arguments=[],
-        args_from_file=False
-    )[0]
+        function="versgetRegistrationCostion",
```
axenteoctavian, line 103: Please fix this function :D

popenta: thanks:) fixed

axenteoctavian (dismissed):

andreibancioiu (approved):

PR #437, opened 2024-07-17
Generate wallet in specific shard

popenta:
For the `mxpy wallet new` command a new **optional** argument `--shard` has been added. When this argument is provided a wallet will be generated in that specific shard.

### DIFF
--- a/multiversx_sdk_cli/cli_wallet.py
+++ b/multiversx_sdk_cli/cli_wallet.py
@@ -34,7 +34,7 @@ WALLET_FORMATS = [
 WALLET_FORMATS_AND_ADDRESSES = [*WALLET_FORMATS, WALLET_FORMAT_ADDRESS_BECH32, WALLET_FORMAT_ADDRESS_HEX]
 
 MAX_ITERATIONS_FOR_GENERATING_WALLET = 100
-CURRENT_SHARDS = [0, 1, 2]
+CURRENT_SHARDS = [i for i in range(NUMBER_OF_SHARDS)]
 
 
 def setup_parser(args: List[str], subparsers: Any) -> Any:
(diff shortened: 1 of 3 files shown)


bogdan-rosianu (dismissed):

multiversx_sdk_cli/cli_wallet.py
```diff
@@ -32,6 +33,9 @@
 
 WALLET_FORMATS_AND_ADDRESSES = [*WALLET_FORMATS, WALLET_FORMAT_ADDRESS_BECH32, WALLET_FORMAT_ADDRESS_HEX]
 
+MAX_ITERATIONS_FOR_GENERATING_WALLET = 100
+CURRENT_SHARDS = [0, 1, 2]
```
andreibancioiu, line 37: Maybe correlate this with:

https://github.com/multiversx/mx-sdk-py-cli/blob/main/multiversx_sdk_cli/constants.py#L15


So that only one constant should be updated in case of such a change.

multiversx_sdk_cli/cli_wallet.py
```diff
@@ -107,8 +112,24 @@ def wallet_new(args: Any):
     format = args.format
     outfile = args.outfile
     address_hrp = args.address_hrp
+    shard = args.shard
+
+    if shard in CURRENT_SHARDS:
+        i = 0
+        while i < MAX_ITERATIONS_FOR_GENERATING_WALLET:
```
andreibancioiu, line 119: Can also be a for from 0 to `max iterations`, with a break within (as you did now).

andreibancioiu (approved):

multiversx_sdk_cli/cli_wallet.py
popenta: Done

multiversx_sdk_cli/cli_wallet.py
popenta: changed to `for` loop

bogdan-rosianu (approved):

andreibancioiu (approved):

PR #441, opened 2024-09-16
Marked `mxpy-up.py` as deprecated

popenta:
Marked `mxpy-up.py` as deprecated and added a required flag when running `mxpy-up.py`.

### DIFF
--- a/mxpy-up.py
+++ b/mxpy-up.py
@@ -21,8 +21,14 @@ def main():
     parser.add_argument("--from-branch", help="use a branch of multiversx/mx-sdk-py-cli")
     parser.add_argument("--not-interactive", action="store_true", default=False)
     parser.add_argument("--verbose", action="store_true", default=False)
+    parser.add_argument("--ignore-deprecation", action="store_false", default=True, help="'mxpy-up.py' is obsolete, install using 'pipx': https://docs.multiversx.com/sdk-and-tools/sdk-py/installing-mxpy/#install-using-pipx")
     parser.set_defaults(modify_path=True)
     args = parser.parse_args()
+
+    if args.ignore_deprecation:
+        logger.warning("'mxpy-up.py' is deprecated. Check out the documentation on how to install using `pipx`: https://docs.multiversx.com/sdk-and-tools/sdk-py/installing-mxpy/#install-using-pipx.")
+        logger.warning("To install using 'mxpy-up.py' set the `--ignore-deprecation` flag.")
+        return
 
     exact_version = args.exact_version
     from_branch = args.from_branch
@@ -55,6 +61,8 @@ Windows support is limited and experimental.
 
     if interactive:
         guide_system_path_integration()
+
+    logger.warning("Installing `mxpy` using `mxpy-up.py` is deprecated. Check out the documentation on how to install using `pipx`: https://docs.multiversx.com/sdk-and-tools/sdk-py/installing-mxpy/#install-using-pipx")
 
 
 def guard_non_root_user():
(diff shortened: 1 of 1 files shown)


mxpy-up.py
```diff
@@ -21,9 +21,15 @@ def main():
     parser.add_argument("--from-branch", help="use a branch of multiversx/mx-sdk-py-cli")
     parser.add_argument("--not-interactive", action="store_true", default=False)
     parser.add_argument("--verbose", action="store_true", default=False)
+    parser.add_argument("--ignore-deprecation", action="store_false", default=True, help="'mxpy-up.py' is obsolete, install using 'pipx': https://docs.multiversx.com/sdk-and-tools/sdk-py/installing-mxpy/#install-using-pipx")
     parser.set_defaults(modify_path=True)
     args = parser.parse_args()
 
+    if args.ignore_deprecation:
```
andreibancioiu, line 28: Ignore deprecation should allow one to install.

The flag should be required. That is, when the user provides it, can install (but sees warnings). If not provided, the user should not be able to install.

Does it make sense?

popenta: It does make sense, but that's what is does. It's not required in the `parser.add_argument(...)` so we can display the warnings, but the code does not execute if the flag is not provided. It's `False` by default and if it's not provided, the `if` statement executes and the `return` statement stops the installation. If provided, the installation process works and the warning is displayed at the end of execution. Let me know if I misunderstood.

andreibancioiu: Oh, now I see.

Can we somehow convert the `return` on line 31 from a happy return to a failure? So that CI workflows that still use mxpy-up don't see a "success" (since it's a failure due to not providing the flag)?

mxpy-up.py
```diff
@@ -21,9 +21,15 @@ def main():
...
+        logger.warning("'mxpy-up.py' is deprecated. Check out the documentation on how to install using `pipx`: https://docs.multiversx.com/sdk-and-tools/sdk-py/installing-mxpy/#install-using-pipx.")
+        logger.warning("To install using 'mxpy-up.py' set the `--ignore-deprecation` flag.")
```
andreibancioiu, line 30: Should be complete failure (e.g. exception) instead a silent success (return code 0).

mxpy-up.py
andreibancioiu, line 28: `If not` maybe?

mxpy-up.py
popenta: Now raising an exception

mxpy-up.py
popenta: done

mxpy-up.py
```diff
@@ -21,14 +21,14 @@ def main():
...
     args = parser.parse_args()
 
-    if args.ignore_deprecation:
-        logger.warning("'mxpy-up.py' is deprecated. Check out the documentation on how to install using `pipx`: https://docs.multiversx.com/sdk-and-tools/sdk-py/installing-mxpy/#install-using-pipx.")
-        logger.warning("To install using 'mxpy-up.py' set the `--ignore-deprecation` flag.")
-        return
+    logger.warning("'mxpy-up.py' is deprecated. Check out the documentation on how to install using `pipx`: https://docs.multiversx.com/sdk-and-tools/sdk-py/installing-mxpy/#install-using-pipx.")
+
+    if not args.ignore_deprecation:
+        raise Exception("'mxpy-up.py' is deprecated, please install using `pipx`: https://docs.multiversx.com/sdk-and-tools/sdk-py/installing-mxpy/#install-using-pipx. If installing using 'mxpy-up` is mandatory, provide the `--ignore-deprecation` flag.")
```
andreibancioiu, line 31: `is necessary` instead of `is mandatory`? Can also stay as it is :+1:

andreibancioiu (approved):

andreibancioiu (approved):

PR #454, opened 2024-10-24
Added cli validators test in pytest and CI for localnet dependent tests.

cristure:


.github/workflows/test-localnet-tests.yml
```diff
@@ -0,0 +1,55 @@
...
+          
+          python3 -m multiversx_sdk_cli.cli localnet clean --configfile=./multiversx_sdk_cli/tests/testdata/localnet_with_resolution_remote.toml
+          python3 -m multiversx_sdk_cli.cli localnet config --configfile=./multiversx_sdk_cli/tests/testdata/localnet_with_resolution_remote.toml
+          nohup python3 -m multiversx_sdk_cli.cli localnet start --configfile=./multiversx_sdk_cli/tests/testdata/localnet_with_resolution_remote.toml > localnet.log 2>&1 & echo $! > localnet.pid
+          sleep 120
+
+      - name: Test localnet dependent tests
+        run: |
+           pytest -m require_localnet .
+
```
andreibancioiu, line 55: Maybe we should stop the localnet as well?

https://github.com/multiversx/mx-sdk-js-core/blob/main/.github/workflows/test-localnet.yml#L61

andreibancioiu (commented):
We can already remove the older bash files, since they are replaced now.

.github/workflows/test-localnet-tests.yml
cristure: the runner dies immediately after the test if finished from what I know. I don't think there is a need to teardown

multiversx_sdk_cli/tests/test_cli_all.sh
```diff
@@ -3,15 +3,14 @@
...
 testAll() {
     pushd $SCRIPT_DIR
     source ./shared.sh
-    source ./test_cli_contracts.sh && testAll
-    source ./test_cli_dns.sh && testAll
 
     if [ -n "$USE_PROXY" ]; then
-        source ./test_cli_validators.sh && testAll
-        source ./test_cli_tx.sh && testAll
         source ./test_cli_config.sh && testAll
```
popenta, line 14: this can be removed together with the `test_cli_config.sh` file.

multiversx_sdk_cli/tests/test_cli_all.sh
```diff
@@ -3,15 +3,14 @@
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
 USE_PROXY=$1
 
+source ./shared.sh
+source ./test_cli_validators.sh && testAll
```
popenta, line 7: not sure why this is here and not in the `testAll` function

multiversx_sdk_cli/tests/test_cli_all.sh
cristure: removed.

multiversx_sdk_cli/tests/test_cli_all.sh
cristure: removed.

popenta (approved):

andreibancioiu (approved):

PR #455, opened 2024-10-25
Remove relayed V3 related code and tests

popenta:
Removed the related relayed V3 code as it was postponed in the protocol. Since the feature never made it to mainnet, I've decided to bump just the **minor** version of the package, even though it is a breaking change.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "multiversx-sdk-cli"
-version = "9.7.1"
+version = "9.8.0"
 authors = [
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "multiversx-sdk-cli"
-version = "9.7.1"
+version = "9.8.0"
```
andreibancioiu, line 7: Indeed, should not be considered a breaking change, in practice.

andreibancioiu (approved):

danielailie (approved):

PR #456, opened 2024-11-01
Reference newer sdk & add client name

popenta:
Fixed a bug in the sdk (https://github.com/multiversx/mx-sdk-py/pull/132) and propagated the fix to `mxpy`.


danidrasovean (approved):

multiversx_sdk_cli/cli_delegation.py
```diff
@@ -209,7 +210,8 @@ def do_create_delegation_contract(args: Any):
 def get_contract_address_by_deploy_tx_hash(args: Any):
     args = utils.as_object(args)
 
-    proxy = ProxyNetworkProvider(args.proxy)
+    config = get_config_for_network_providers()
+    proxy = ProxyNetworkProvider(url=args.proxy, config=config)
```
andreibancioiu, line 214: In a future PR, we can have a shared function such as `create_network_provider(url: str)`, to hold both these lines etc.

andreibancioiu (approved):

multiversx_sdk_cli/cli_delegation.py
popenta: Indeed, was thinking the same.

PR #481, opened 2025-02-11
Create validator wallet

popenta:
Added functionality to create a validator wallet.
Also added functionality to sign a message using a validator wallet and to verify a messages using a bls validator pubkey.

### DIFF
--- a/multiversx_sdk_cli/cli_validator_wallet.py
+++ b/multiversx_sdk_cli/cli_validator_wallet.py
@@ -38,7 +38,7 @@ def setup_parser(args: list[str], subparsers: Any) -> Any:
     sub.add_argument("--message", required=True, help="the message you want to sign")
     sub.add_argument("--pem", required=True, type=str, help="the path to a validator pem file")
     sub.add_argument(
-        "--validator-index",
+        "--index",
         required=False,
         type=int,
         default=0,
@@ -91,7 +91,7 @@ def create_new_wallet(args: Any):
 
 def sign_message(args: Any):
     path = Path(args.pem).expanduser().resolve()
-    validator_signer = ValidatorSigner.from_pem_file(path, args.validator_index)
+    validator_signer = ValidatorSigner.from_pem_file(path, args.index)
     signed_message = sign_message_by_validator(args.message, validator_signer)
 
     utils.dump_out_json(signed_message.to_dictionary())


multiversx_sdk_cli/cli_wallet.py
```diff
@@ -57,6 +75,12 @@ def setup_parser(args: List[str], subparsers: Any) -> Any:
         "new",
         "Create a new wallet and print its mnemonic; optionally save as password-protected JSON (recommended) or PEM (not recommended)",
     )
+    sub.add_argument(
+        "--validator-wallet",
```
andreibancioiu, line 79: I wonder whether we should have a separate sub-command, instead.

E.g.

```
mxpy wallet ... (for user wallets)
mxpy validator-wallet ...
```

Especially since other sub-sub-commands of `mxpy wallet` do not make sense for this kind of wallets (e.g. _bech32_, _convert_ etc.). Actually, some _convert_ functionality could be useful (e.g. PEM to raw hex), but, overall, maybe we can have them separately.

:thought_balloon:

multiversx_sdk_cli/sign_verify.py
```diff
@@ -48,3 +69,10 @@ def to_dictionary(self) -> Dict[str, str]:
 def sign_message(message: str, account: IAccount) -> SignedMessage:
     signature = account.sign_message(Message(message.encode()))
     return SignedMessage(account.address.to_bech32(), message, signature.hex())
+
+
+def sign_message_by_validator(message: str, validator: ValidatorSigner) -> SignedMessage:
+    message_computer = MessageComputer()
+    serialized_message = message_computer.compute_bytes_for_signing(Message(message.encode()))
```
andreibancioiu, line 76: For validator wallets, this is not e regular use-case. In the few cases when validator wallets are required to sign a message, they sign the plain variant (raw bytes). Not this special enveloping:
 - https://github.com/multiversx/mx-sdk-py/blob/main/multiversx_sdk/core/message.py#L39

andreibancioiu: As far as I know (see validators CLI of mxpy).

multiversx_sdk_cli/cli_wallet.py
popenta: Moved to a separate sub-command

multiversx_sdk_cli/sign_verify.py
popenta: Fixed

andreibancioiu (approved):

danielailie (approved):

PR #483, opened 2025-02-19
Use native auth client from sdk-py

popenta:


danielailie (approved):

requirements.txt
```diff
@@ -8,4 +8,5 @@ ledgercomm[hid]
 rich==13.3.4
 argcomplete==3.2.2
 
-multiversx-sdk[ledger]==1.1.0
+# multiversx-sdk[ledger]==1.1.0
```
andreibancioiu, line 11: We should not forget about this (should be fixed right before release).

andreibancioiu (approved):

requirements.txt
popenta: indeed, it will

PR #488, opened 2025-02-24
Use single index argument for each wallet

popenta:


danielailie (dismissed):

multiversx_sdk_cli/cli_shared.py
```diff
@@ -177,10 +165,10 @@ def add_wallet_args(args: list[str], sub: Any):
         help="🔐 bool flag for signing transaction using ledger",
     )
     sub.add_argument(
-        "--ledger-address-index",
+        "--index",
```
andreibancioiu, line 168: This is somehow ambiguous. Maybe `--sender-index`?

For _guardian_ & _relayer_ is more obvious what index means.

andreibancioiu (approved):
Can also be:

```
--sender-wallet-index
--guardian-wallet-index
--relayer-wallet-index
```

A bit longer, but precise.

multiversx_sdk_cli/cli_shared.py
popenta: renamed as follows:
```
--sender-wallet-index
--guardian-wallet-index
--relayer-wallet-index
```

andreibancioiu (approved):

danielailie (approved):

PR #494, opened 2025-03-14
Add specific prefix for address arguments

popenta:
Whenever an address is provided as an smart contract call argument, the `addr:` prefix should be used.
e.g.
```
mxpy contract call ... --arguments addr:erd1...
```

### DIFF
--- a/multiversx_sdk_cli/contracts.py
+++ b/multiversx_sdk_cli/contracts.py
@@ -267,6 +267,10 @@ class SmartContract(BaseTransactionsController):
             elif arg.startswith(ADDRESS_PREFIX):
                 args.append(AddressValue.new_from_address(Address.new_from_bech32(arg[len(ADDRESS_PREFIX) :])))
             elif arg.startswith(MAINCHAIN_ADDRESS_HRP):
+                # this flow will be removed in the future
+                logger.warning(
+                    "Address argument has no prefix. This flow will be removed in the future. Please provide each address using the `addr:` prefix. (e.g. --arguments addr:erd1...)"
+                )
                 args.append(AddressValue.new_from_address(Address.new_from_bech32(arg)))
             elif arg.startswith(get_address_hrp()):
                 args.append(AddressValue.new_from_address(Address.new_from_bech32(arg)))
--- a/multiversx_sdk_cli/contracts.py
+++ b/multiversx_sdk_cli/contracts.py
@@ -36,6 +36,8 @@ HEX_PREFIX = "0x"
 FALSE_STR_LOWER = "false"
 TRUE_STR_LOWER = "true"
 STR_PREFIX = "str:"
+ADDRESS_PREFIX = "addr:"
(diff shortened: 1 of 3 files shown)


multiversx_sdk_cli/contracts.py
```diff
@@ -262,6 +264,10 @@ def _prepare_args_for_factory(self, arguments: list[str]) -> list[Any]:
                 args.append(BytesValue(self._hex_to_bytes(arg)))
             elif arg.isnumeric():
                 args.append(BigUIntValue(int(arg)))
+            elif arg.startswith(ADDRESS_PREFIX):
+                args.append(AddressValue.new_from_address(Address.new_from_bech32(arg[len(ADDRESS_PREFIX) :])))
+            elif arg.startswith(MAINCHAIN_ADDRESS_HRP):
+                args.append(AddressValue.new_from_address(Address.new_from_bech32(arg)))
```
andreibancioiu, line 270: For this flow, perhaps add a comment that we'll remove it in the future?

multiversx_sdk_cli/contracts.py
```diff
@@ -36,6 +36,8 @@
 FALSE_STR_LOWER = "false"
 TRUE_STR_LOWER = "true"
 STR_PREFIX = "str:"
+ADDRESS_PREFIX = "addr:"
+MAINCHAIN_ADDRESS_HRP = "erd"
```
andreibancioiu, line 40: Can we remove this constant, and directly use the field from `LibraryConfig` below?

multiversx_sdk_cli/contracts.py
andreibancioiu: Maybe even display a warning?

pyproject.toml
```diff
@@ -26,7 +26,7 @@ dependencies = [
   "ledgercomm[hid]",
   "rich==13.3.4",
   "argcomplete==3.2.2",
-  "multiversx-sdk[ledger]==1.1.0"
+  "multiversx-sdk[ledger]==1.2.0b0"
```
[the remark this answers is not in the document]
popenta: done

multiversx_sdk_cli/contracts.py
popenta: hmm, I am not sure about this. What if somebody changes the `LibraryConfig`?

andreibancioiu: Generally speaking, since `mxpy` is the _end application_ (not a library), only us can be that somebody.

If someone changes the HRP in mxpy's config (legacy) or via the CLI (I think not available on all commands for now), she will also get the benefice of this legacy flow (if you would remove the constant and directly use the field from library config).

Having "erd" only for this legacy flow might be misleading & might lead to additional questions / support. Even if the flow is legacy.

andreibancioiu (approved):

danielailie (approved):

PR #509, opened 2025-03-31
Fetch HRP from network if possible

popenta:


multiversx_sdk_cli/cli_shared.py
```diff
@@ -302,8 +303,39 @@ def prepare_account(args: Any):
...
+
+    if hasattr(args, "proxy") and args.proxy:
+        hrp = _get_hrp_from_proxy(args)
+    elif hasattr(args, "api") and args.api:
+        hrp = _get_hrp_from_api(args)
+
+    if hrp:
+        return hrp
+
+    return config.get_address_hrp()
```
axenteoctavian, line 318: There are some github warnings for this

axenteoctavian (commented):
This implementation works, but for some commands you call /network/config twice (once for hrp and once for chainID). Maybe you can create a task to only call it once per request, if it's not a quick fix

multiversx_sdk_cli/cli_shared.py
popenta: fixed

popenta:
> This implementation works, but for some commands you call /network/config twice (once for hrp and once for chainID). Maybe you can create a task to only call it once per request, if it's not a quick fix

That is a good idea. Will probably use a cache and will get the chainID from the cache. Will do it in a future PR since quite a few changes are needed.

axenteoctavian (approved):

andreibancioiu (approved):

PR #513, opened 2025-04-09
Fix displaying transaction on network

popenta:


pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "multiversx-sdk-cli"
-version = "10.1.0"
+version = "10.1.1"
```
andreibancioiu, line 7: Debatable - maybe minor version, since it's a theoretical breaking change in the output format.

multiversx_sdk_cli/cli_output.py
```diff
@@ -69,7 +69,7 @@ def build(self) -> dict[str, Any]:
             output["contractAddress"] = contract_address
 
         if self.transaction_on_network:
-            transaction_on_network_dict = transaction_on_network_to_dictionary(self.transaction_on_network)
+            transaction_on_network_dict = self.transaction_on_network.raw
```
andreibancioiu, line 72: Good, now we return the data as it comes from the API (no change - except of the field omissions below) :+1:

andreibancioiu (approved):

pyproject.toml
andreibancioiu: Let's have it as patch, as it is :+1:

danielailie (approved):

PR #519, opened 2025-05-27
Environment config

popenta:


danielailie (dismissed):

multiversx_sdk_cli/cli_env.py
```diff
@@ -0,0 +1,149 @@
...
+    )
+    _add_name_arg(sub)
+    sub.add_argument(
+        "--template",
+        required=False,
+        help="an environment from which to create the new environment",
+    )
+    sub.set_defaults(func=new_env)
+
+    sub = cli_shared.add_command_subparser(subparsers, "env", "get", "Gets an env value from the active environment.")
```
andreibancioiu, line 41: Maybe `get-value` instead of `get` and `set-value` instead of `set`? Just an opinion.

andreibancioiu (approved):

PR #522, opened 2025-06-04
Added get commands

popenta:
Added a couple of get commands for the most common use cases.

Added a command to get an account from the network: `mxpy get account`.
Added a command to get the storage of an account from the network: `mxpy get keys`.
Added a command to get a key from an account's storage: `mxpy get storage-entry`.
Added a command to get a token of an account: `mxpy get token`.
Added a command to get a transaction from the network: `mxpy get transaction`.
Added a command to get the network config: `mxpy get network-config`.
Added a command to get the network status: `mxpy get network-status`.

### DIFF
--- a/CLI.md.sh
+++ b/CLI.md.sh
@@ -207,7 +207,7 @@ generate() {
 
     group "Get" "get"
     command "Get.Account" "get account"
-    command "Get.Keys" "get keys"
+    command "Get.Storage" "get storage"
     command "Get.StorageEntry" "get storage-entry"
     command "Get.Token" "get token"
     command "Get.Transaction" "get transaction"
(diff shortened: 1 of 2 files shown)


CLI.md.sh
```diff
@@ -193,6 +193,24 @@ generate() {
...
+    command "Address.Get" "address get"
+    command "Address.Set" "address set"
+    command "Address.Set" "address delete"
+    command "Address.Switch" "address switch"
+    command "Address.Remove" "address remove"
+    command "Address.Reset" "address reset"
+
+    group "Get" "get"
+    command "Get.Account" "get account"
+    command "Get.Keys" "get keys"
```
ovidiuolteanu, line 210: I would change "keys" to "storage" to keep it consistent with the other "storage-entry" term and docs, otherwise it's confusing.

popenta: done. will re-generate CLI.md before releasing the new version.

ovidiuolteanu (approved):

PR #533, opened 2025-06-25
Rename env to config-env

popenta:


multiversx_sdk_cli/config_env.py
```diff
@@ -46,22 +46,59 @@ def get_defaults() -> dict[str, str]:
...
 
 
 @cache
 def get_explorer_url() -> str:
-    return get_value("explorer_url")
+    """
+    Returns the explorer URL for the active environment.
+    If not set, it returns an empty string.
+    """
+    data = read_env_file()
```
danielailie, line 81: can you extract this in a method and use it in all the other functions also?

andreibancioiu (approved):

danielailie (approved):

PR #565, opened 2026-04-14
Fix pipx install and align CI workflows across systems

BubuMVX:
## Summary

This PR fixes the `pipx install` Git URLs used in CI so pull requests from forked repositories and feature branches install the correct source, even if the branch name contains special characters.

It also aligns the related GitHub Actions workflows across platforms.

## Changes

- update the Ubuntu and macOS `pipx install` workflows to build the Git URL from the current repository and branch
- add the same repository/branch-based install logic to the Windows `pipx install` workflow
- include Windows in the main build matrix
- include Windows in the localnet test matrix
- replace the old Windows-specific build workflow with a simple `pipx install` workflow and smoke test

## Why

The previous `pipx install` commands were hardcoded to the upstream repository, which could install the wrong code when validating a pull request coming from a fork.

They could also fail with branch names containing special characters, such as `/`, because the Git URL was not properly formed for those cases.

With this change, CI installs directly from the repository and branch associated with the pull request.


popenta:
We are not allowing installing directly from the forked repository as it may be a serios security issue. Will close the PR.

